PR #18, opened 2016-06-26
Fix bad formatting and change version numbers to 3.8 in INSTALL_WIN.md

luciusmagn:
and change lines 88-95 to take CMakeLists.txt changes in account


bvdberg:
merged!

PR #22, opened 2017-04-08
Windows port update

luciusmagn:
This pull-request updates CMakeLists.txt to use g++ on Windows/Cygwin because of a bug with clang++ and libstdc++'s __float128 type. It also updates the instructions for Windows installation to be up-to-date with Clang/LLVM 4.0 and building LLVM with CMake instead of the configure script. It also adds Cygwin to targets


bvdberg:
merged!

PR #35, opened 2018-10-26
[FIX] LiteralAnalyser: Missing shift operations causes compiler crash.

lerno:
Added code for handling shr/shl operators in LiteralAnalyser. The previous default made them return a 1 bit result :(

Added warnings for lhs < 0, rhs < 0, and shift where shift was wider than lhs bit width.
Should probably add warning for lhs overflow too. But that requires a discussion. Added tests, one to reproduce the problem, three for the warnings.

I also added a missing include for when C2Parser has debug on.

Minor refactoring: since getting lhs/rhs was common, I extracted those values into locals.

### DIFF
--- a/c2c/Analyser/LiteralAnalyser.cpp
+++ b/c2c/Analyser/LiteralAnalyser.cpp
@@ -119,6 +119,7 @@ void LiteralAnalyser::check(QualType TLeft, const Expr* Right) {
     StringBuilder tname(128);
     TLeft->DiagName(tname);
     const Limit* L = getLimit(availableWidth);
+
     checkWidth(availableWidth, L, Right, tname);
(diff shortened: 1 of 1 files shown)


bvdberg:
Nice one! Well done.
I merged your changes. After that I promoted the warnings to error severity. I've done that with most warnings. The only thing in C2 that is a warning are unused variables/functions/etc.

When I parse the original fragment now, it gives the error, but after that there's still the assert. I think we need to refactor the checkLiterals() function to check if both sides had any errors.. Won't make it prettier..

bvdberg:
I pushed a quick patch just before you submitted your more thorough one. In the merged, I've used your solution. Much better..

PR #36, opened 2018-10-26
[LANG-CHANGE] Update precedence rules, flattening the precedence hier…

lerno:
…archy and increasing precedence of bitwise operators.

I did the (almost) simplest possible change here, adding () in as simple way as possible. A nicer solution would have been to actually check if the parent expression already was surrounded by () when printing to C.

Note that I haven't looked at anything but the AST and the C output. There might be subtle things that I have overlooked. Testcases are updated and a new test case for some of operators have been added.

### DIFF
--- a/c2c/AST/Expr.cpp
+++ b/c2c/AST/Expr.cpp
@@ -543,6 +543,34 @@ const char* BinaryOperator::OpCode2str(clang::BinaryOperatorKind opc_) {
     assert(0);
 }
 
+bool BinaryOperator::requiresParensForC() const
+{
+    switch (getOpcode()) {
+    case BO_Mul:
+    case BO_Div:
+    case BO_Rem:
+    case BO_Add:
+    case BO_Sub:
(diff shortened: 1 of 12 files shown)


bvdberg:
This is a pretty big change (not codewise, but language wise). Could you perhaps show the consequences by some examples?

bvdberg:
The patch itself looks in good shape. I only want to merge it when I've also updated the documenation..

lerno:
@bvdberg how should I send the examples? Write here? The forum? Wiki?

bvdberg:
I think the best way would be to change the documentation (in c2lang/docs
from head)...or else just write them in email and I'll convert to doc site

lerno:
I added to the wiki now @bvdberg

bvdberg:
I've given the Wiki and c2lang.org/docs/ an update. I've also restricted the Wiki access. Apparently the default on github is that everyone can edit.

PR #39, opened 2018-10-27
[FIX] Operators < > >= <= == != ^ | & and comma should now evaluate c…

lerno:
…orrectly in the literal analysis.

I made the decision that when comparing two values using relational operations, the result is the max size of the two operands and 0 or 1.

Some refactoring. This also fixed the error I accidentally introduced in my pull req (I corrected it but then you had already integrated my changes from the previous pull req. See the BO_Mult case: lhs twice...)


bvdberg:
merged

Issue #43, opened 2018-10-29
Completely remove dependency on Clang's lexer


lerno:
1. The lexer parses a lot of stuff that C2 shouldn't, this makes the code rather unclear at places.
2. The lexer has a firm idea about how to lex certain tokens, like "@" which might complicate parsing.
3. We have full C pre-processing. I don't think this is a good thing.
4. Keywords need to be defined in the Clang source :-[
5. Code requires reading up on a lot of Clang code to see how things work, rather than just the c2compiler code.

bvdberg:
I think replacing the Clang preprocessor/Lexer with a custom version seems nice, but it doesn't really bring anything. Language wise it changes nothing. Also Clang's lexer generates the SourceLocation's for each token. These are also used in Clang's DiagnosticsEngine. So either both need to be replaced, or a replacement Lexer would have to generate compatible ones.

I think that before doing this, we need to do a *lot* more analysis, otherwise we run the risk of ending up in a broken state..

lerno:
It's definitely not lightly done, and I'm *absolutely not* suggesting that this is done in a single sweep. Rather, I suggest that a roadmap is drawn up to lessen direct dependency on Clang. Refactorings are done piecemeal until the dependency on Clang is completely gone. This is a slow, baby step by baby step, change.

I completely understand that you might have misunderstood me. I'm used to working in legacy code bases where it's important that the code is constantly improved, refactored etc, BUT it must always be in working condition. So changes are small, chipping away at the old and building the new step by step.

Tools in the toolbox:
1. Making wrappers
2. Relying on fewer functions
3. Auditing what parts of each class is really used.
4. Copy-paste code from Clang and use the copies instead of the normal clang classes. Then refactor these copies small pieces at a time (if necessary)

lerno:
Just a quick update. I'm working on the 4th approach now. Making a copy of Clang's Lexer & Diagnostics + all the necessary files that are attached to it (they are MANY)

Once the lexer is part of the "normal" C2C source:

1. Remove clang as library, just depend on a vanilla llvm.
2. Strip parts not needed for the lexing / preprocessing of C2 (that's *a lot*)
3. Eventually replace the lexer and streamline the diagnostics

Already at (1), we can do any changes to keywords etc directly in the C2C source, which is a huge plus for me.

lerno:
I managed to extract all the code to have it compile (with only name / path changes). Have a look when you have time: https://github.com/lerno/c2compiler/tree/no_clang

[...]

PR #46, opened 2018-11-01
No clang

lerno:
Following #43, I think that this extracted piece of Preprocessor / Lexer / Diagnostics from Clang is suffiently small to be added to C2C. It might be better to rebase everything in a single commit. I'm leaving it in multiple commits so that it's possible to track how I "peeled the onion". The only point where it doesn't compile is  commit 7a6c86e


lerno:
Some other things: this branch makes it pretty easy to work with new keywords directly and change details of the lexing. I think eventually we should have a lexer written from scratch for C2, but this one is now stripped of most things. The preprocessor remains, but is limited to if/ifndef/normal macros. #include is gone as is all non standard directives. The Clang preprocessor and lexer is very flexible since they have to support different keywords, directives etc depending on compiler arguments. In addition, the modules is an EXTREMELY invasive addition to Clang, forcing a lot of added complexity and necessary flexibility that can be eliminated with C2.

I've wanted to add some other language features, but since this change spans ALL files, I've based those changes off this branch instead.

lerno:
Diagnostics should be reviewed, removing everything that shouldn't be there and setting the right defaults for everything, but since I haven't used it much I didn't want to poke around. It was very easy to break things.

PR #56, opened 2018-11-06
Ir improvements

lerno:
Some additions to the IR generation and clarified state in analysis.


lerno:
Note that this one contains the contents of the earlier pull req to remove dependency on Clang. I've waited about a week to see if the #55 (#46) would be accepted, to know which branch I should work from. 

After working extensively on stripping the Clang code (it's now close to a minimum), I don't see how it's reasonable to go back, there are so many weird C things one gets by piggybacking on Clang. Unless the idea is a C 1.1 – where most of C is still there, it's not going to be possible to use Clang as a lexer in the long run.

So I'm making the assumtion that we go for that change.

PR #57, opened 2018-11-06
No clang w include

lerno:
This alternative checkin skips the last few commits, stopping at a point where include is still fully working. It is an alternative to #55 and #46


bvdberg:
This branch differs a lot from the clang_parts_to_c2c. More than 40.000 lines of code. Did you really remove all that code before?

lerno:
@bvdberg Yes! Pulling out the include directive allowed me to remove much of the header searching etc. After that I also found other things (not related to include) that I could remove, so that should be correct. You can compare with the no_clang branch, there you can see the additional commits that encompass those additional 40k of deleted lines. I spent a long time on this.

bvdberg:
Indeed, Maybe you can still keep that patch on a branch for now and we'll get to work on a macro system asap..

lerno:
The cleanup is not strictly necessary I think. We can replace it when we replace the lexer some time in the future.

bvdberg:
I'm merging it.. just updating the INSTALL.md as well..

bvdberg:
doing a final clean build (including llvm) so takes around an hour.

bvdberg:
The end of an era. C2 no longer depends on Clang!

PR #59, opened 2018-11-07
Some IR work, switching from plain iteration.

lerno:
- Changed to use foreach loops - skip this one if you want old school loops.
- Forced to add a isFloat() to QualType
- Added explicit state to the FileAnalyser. Note that this is not actually adding any behaviour, but it becomes very clear what can happen in what state. It was very hard to tell from just reading the code. It can be excluded.
- Fixed some unary IR (not/neg) emits.
- Added a break-continue stack (I actually started on the for loop, but it's not included in this checkin)

### DIFF
--- a/c2c/AST/Expr.h
+++ b/c2c/AST/Expr.h
@@ -98,6 +98,11 @@ public:
     bool hasImpCast() const {
         return getImpCast() != BuiltinType::Void;
     }
+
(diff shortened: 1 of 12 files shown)


lerno:
Found the branch I should have used which has some additional changes. Closing this one.

PR #60, opened 2018-11-07
Some improvements to IR. Also clarified state in AST analysis.

lerno:
- Changed to use foreach loops - skip this one if you want old school loops.
- Forced to add a isFloat() to QualType
- Added explicit state to the FileAnalyser. Note that this is not actually adding any behaviour, but it becomes very clear what can happen in what state. It was very hard to tell from just reading the code. It can be excluded.
- Added a break-continue stack (I actually started on the for loop, but it's not included in this checkin)
- Fixed some unary IR emits (~, -, !, +).
- Add implicit casting for binary expressions. Probably should already be done in AST analysis instead.
- * - + should now work both for floats and ints
- xor/or/and works for ints

### DIFF
--- a/c2c/AST/Expr.h
+++ b/c2c/AST/Expr.h
@@ -98,6 +98,11 @@ public:
     bool hasImpCast() const {
         return getImpCast() != BuiltinType::Void;
     }
+
+    bool isFloat() const {
+        return QT.isFloatType();
+    }
+
     BuiltinType::Kind getImpCast() const {
         return static_cast<BuiltinType::Kind>(exprBits.ImpCast);
     }
--- a/c2c/AST/Type.cpp
+++ b/c2c/AST/Type.cpp
@@ -63,6 +63,23 @@ bool QualType::isIntegerType() const {
     }
     return false;
 }
+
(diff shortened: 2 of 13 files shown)


bvdberg:
There are 2 branches, some_IR_work and ir_improvements, what is the difference?
I've looked at some_IR_work and that seems to be in the right shape (except no new tests).
when adding new analyser/generation functionality, i except some new unit tests. This way is also very easy to see what has been added.

One other thing, the current codebase uses 'Type* var' instead of 'Type *var'. I personally also prefer
this, since in my opinion, the type is 'Type*'.

bvdberg:
Did you know that you can run unit-tests individually by running:
```
./build/tools/tester test/Scope
./build/tools/tester test/CGenerator/hello_world.c2t
```

speeds things up a bit ;)

lerno:
I actually write a small test case and compile manually since the tester framework swallows all debug code :) 

I usually do `make tests` which I don't think have that ability to pinpoint individual tests?

bvdberg:
`make tests` actually does (inside build/)  `tools/tester ../test`

PR #62, opened 2018-11-08
Fixes chained struct code like foo.getA().getB().

lerno:
The previous code could only handle a single call. The fix instead introduces a call stack. I also split the analysis of "structMember" to improve readability a little bit.

### DIFF
--- a/c2c/Analyser/FunctionAnalyser.cpp
+++ b/c2c/Analyser/FunctionAnalyser.cpp
@@ -39,6 +39,7 @@ using namespace c2lang;
 //#define ANALYSER_DEBUG
 
 #ifdef ANALYSER_DEBUG
+#include "Utils/color.h"
 #include <iostream>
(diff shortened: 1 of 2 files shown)


bvdberg:
Can you add some unit tests (also with C-code generation) as well?

bvdberg:
something like:
```
module test;

type A struct {
}

func u32 A.run(A* a) {
    return 10;
}

type B struct {
    A a;
}

func A* B.getA(B* b) {
    return &b.a;
}

type C struct {
    B b;
}

func B* C.getB(C* c) {
    return &c.b;
}

func void test1() {
    C c;
    u32 n = c.getB().getA().run();
}
```

bvdberg:
It's amazing that the C-code generation works without changes! So
`c.getB().getA().run()'

becomes:
`test_A_run(test_B_getA(test_C_getB(&c)));`

sweet!

bvdberg:
No matter, I'll add some tests and merge...

PR #70, opened 2018-11-18
Subunion

lerno:
Fixes anonymous unions and issues warnings on union init without designators. Note that this does not mean that struct init is completely working, just that we're closer to having it work.

I've split up the pull req into 3 commits to make it more readable.

1. First commit simply refactors the code a bit into smaller pieces.
2. Second commit fixes anonymous unions.
3. Third commit adds the error reporting for unions not using designators for init.


bvdberg:
Indeed still work in progress, but an improvement, so merged

PR #72, opened 2018-11-20
Further cleanup

lerno:
Further removals from the Clang code + one adjustment of the precedence rules so that they hold in macro expressions as well (something I overlooked when I first did this)


bvdberg:
looks good, nothing to comment, just merged..

PR #75, opened 2018-11-26
Bigint example

lerno:
A simple big int example. Just for test, not for serious use.


bvdberg:
merged (did minor style change)

PR #79, opened 2018-12-01
Fix sizeof. Also inlines many sizes variable sizes directly. However,…

lerno:
… structs are skipped as exact sizes are unclear at this point of time. A bug was found, where – for example – "Foo a = Foo;" is allowed. This requires some extra code to distinguish between typevalues and types of expressions. Since this bug already existed, I did not want to do the extra work in this commit, however, I've updated the expected error.

### DIFF
--- a/c2c/AST/Decl.cpp
+++ b/c2c/AST/Decl.cpp
@@ -328,6 +328,21 @@ Decl* StructTypeDecl::find(const char* name_) const {
     return findFunction(name_);
 }
 
+Decl* StructTypeDecl::findMember(const char* name_) const {
+    // normal members
+    for (unsigned i=0; i<numMembers(); i++) {
+        Decl* D = members[i];
+        if (strcmp(D->getName(), name_) == 0) return D;
+        if (D->hasEmptyName()) {      // empty string
+            assert(isa<StructTypeDecl>(D));
+            StructTypeDecl* sub = cast<StructTypeDecl>(D);
+            D = sub->find(name_);
+            if (D) return D;
+        }
+    }
+    return nullptr;
+}
+
 FunctionDecl* StructTypeDecl::findFunction(const char* name_) const {
     // struct-functions
     for (unsigned i=0; i<numStructFunctions(); i++) {
--- a/c2c/AST/Decl.h
+++ b/c2c/AST/Decl.h
@@ -351,6 +351,7 @@ public:
     unsigned numStructFunctions() const { return structTypeDeclBits.numStructFunctions; }
     Decl* getMember(unsigned index) const { return members[index]; }
     Decl* find(const char* name_) const;
+    Decl* findMember(const char* name_) const;
     FunctionDecl* findFunction(const char* name_) const;
     int findIndex(const char*  name_) const;
     void setOpaqueMembers();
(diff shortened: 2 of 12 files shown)


bvdberg:
I see that the Parser now indeed uses the Capital case/not to determine it it's a type. There's one corner case
here (just impacts other stuff) and that's external C libraries. External C libraries are allowed to have types with lower cased characters. So that's very nasty.

I also ran into this issue with the stat() function in libC. There a function stat() and a struct stat. Hmmm..
What I still have to try is to change struct stat -> struct Stat and use an attribute (c-name='stat'). That way
we can maintain the  spelling and keep  this code.

Why did the following test change: test/Functions/struct_functions/static_member.c2
The error seemed ok, but now it's 'i32 cannot be assigned to value'

lerno:
My suggestion was that C structs are imported as _C_<struct_name> into C2 so:

1. struct foo in plain C is imported as _C_foo. This means our naming standard is retained (the _C_ prefix is interpreted an upper case struct)
2. function Foo in plain C is imported as _c_Foo. Again naming standard is retained.
3. Since we disallow _C_ as prefix for C2 structs, this is no issue.
4. This also fixes the problem with function / struct in C that you mention.

The only downside is for C functions that follow the name standard. I suggest that those are imported with prefix, but aliased to their normal name.

So:

C:

1. struct Foo {  … }
2. struct bar { … }
3. void Foo( … )
4. void bar( … )

Imported into C2 as:

1. Foo and _C_Foo
2. _C_bar
3. _c_Foo
4. bar and _c_bar

This should resolve all issues.

For the less nice ”_C_bar”, it’s easy to alias it with a good name.

/C

lerno:
@bvdberg re-read my suggestion.

printf would be available as is without change. The prefix is only for the ones not matching our conventions.

lerno:
Any other questions on this one @bvdberg ? Next after this is tackling the struct packing / calculations

lerno:
I removed one diff in indent to make the diff even smaller. Will there be time to merge this?

lerno:
At this point, everything except structs will inline the sizes. This was simpler than sending it to C actually. You should have several examples already, but sure, I can add more.

lerno:
I realized you're right, the C-gen was not there. Added now.

bvdberg:
merged after adding some tests and fixing some static struct stuff..

lerno:
Would you mind explaining what you did @bvdberg? Looks like you changed a lot compared to my implementation.

bvdberg:
I did some refactoring indeed. The base was to get the unit test test/Expr/member_expr.c2 running correctly. Most changes have nothing to do with @lerno 's patch, but issues already present.

- We did not keep track of whether a function was a static-struct func or not. So some code moved from FunctionAnalyser to FileAnalyser (like _isStaticStructFunc_() and it's utility function _getStructDecl_)
- To fix the member_expr unit test, the analyser has to keep track of whether or not to allow static access to member. For sizeof(x), it's allowed. Static struct functions are also allowed. But other static use of nonstatic struct members is not. The variable **allowStaticMember** is used for this.
- The analyseStructMember broke some unit-tests by allowing too much. This function was refactored, based on the unit tests.
- added some diagnostic messages for accessing non-static members
- added unit tests

Issue #50, opened 2018-11-03
unary + is not pruned from AST.


lerno:
There is no need to leave the unary + in the AST, this just makes analysis a bit more difficult. Either prune it in the lexer (when doing the lexer) or prune it in the AST during parsing.

bvdberg:
agreed

lerno:
Note that the + operator actually has some function in C:
https://stackoverflow.com/questions/6637005/what-is-the-purpose-of-the-unary-operator-in-c

Three possibilities:
1. Remove it from the grammar
2. Make it a noop (no int promotion)
3. Use it as C does (int promotion, error on pointers with +)

bvdberg:
Option 1 is definitely the best IMHO. I've also never seen it used anywhere.

lerno:
The point of unary plus is that you can write something like

```
add[] = {
  +1,
  -1,
  -1,
  +1 };
```

Retaining visual symmetry. I would prefer keeping it for that reason.

bvdberg:
If it's not in the way and just removed from the AST, fine by me.

lerno:
Thank you. I'll close this one then.

PR #86, opened 2018-12-06
#50 Prune unary plus from AST, making it a complete no-op.

lerno:
This is a version of unary plus pruning that completely makes it a no-op by skipping over it in AST parsing already. See discussion around #50 on how this differs from C.

Note that this change does not affect normal binary operations.


bvdberg:
See comments in issue #50

bvdberg:
tested and merged, good job!

PR #90, opened 2018-12-10
Cut away LangOptions and unused functionality

lerno:
Continuing my quest to cut away pieces of Clang that isn't actually used by the Lexer when C2 invokes it.


bvdberg:
merged

PR #93, opened 2018-12-19
Better about message

Dandigit:
Modified the about message to be grammatically correct and more readable.

Here's the modified message; I've updated `main.cpp` to use it:
```
The C2 Compiler by Bas van den Berg 

C2 is a programming language that aims to keep the good parts 
of C and remove/improve the bad parts. C2 provides stricter 
and cleaner syntax, better compilation times, great tooling 
and a smart build system among other things. 

C2 aims to be the direct successor to C, replacing it in  
domains currently dominated by C, like bootloaders, kernels, 
drivers and system-level tooling. 

C2 is based on LLVM and Clang. 
See c2lang.org for more information. 
```


bvdberg:
thx

PR #102, opened 2019-01-29
Trimming down Clang usage, mostly eliminating frameworks from headersearch

lerno:
The only thing untested here is checking whether #include directives still work as expected, otherwise it's only clang internal code that's touched.


lerno:
There are unfortunately some larger diffs as I ran code reformat on some files as the formatting was poor and hard to read.

bvdberg:
Are C macro's still working?

bvdberg:
I did some quick performance tests and it seems this branch is a bit faster... :) :)

PR #103, opened 2019-01-31
Static build options

lerno:
Build options are global from the invocation, but is passed around C++ style as a dependency. Make this global property explicit instead. Any options that are not global should obviously be passed into C2Builder, but build options indeed seem to be fully global.

### DIFF
--- a/c2c/Builder/C2Builder.cpp
+++ b/c2c/Builder/C2Builder.cpp
@@ -160,10 +160,9 @@ public:
 }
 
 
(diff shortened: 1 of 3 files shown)


lerno:
Feel free to reject this pull req. If it is kept, I suggest build options to be moved into a separate file.

bvdberg:
Yeah I prefer moving them to a separate file... When we can we should avoid globals..

PR #109, opened 2019-08-15
Update INSTALL.md

paulie-g:
On current Arch (and that's the only kind of Arch ;), libtinfo is included with ncurses.


bvdberg:
Indeed. It must keep working on Arch :) thx!

PR #119, opened 2020-12-01
Proposed changes to work with LLVM11

lerno:
See if these are correct.


bvdberg:
Yes, they appear to work. I did you the clang mainline way, so instead of Code.str(), I used std::string(Code).
Nice one with the LoadInst! Already merged to master

PR #134, opened 2024-12-18
Partial port to Darwin arm64

chqrlie:
bootstrap:
* keep intermediary target **c2c_bootstrap2**
* use dynamic compiler detection for gcc/clang and target architecture
* adjust warning flags for gcc/clang
* define `fallthrough` according to compiler version
* use `#ifdef` to select architecture dependent definitions
* fix `target_info_arch_names` name as `arm64` instead of `arm_64`
* add **globals** to generate C library constants and structures

common/target_info.c2:
* add `Info.define` for C library source preprocessing
* use `arm64` instead of `arm_64` in `arch_names`

generator/:
* generate `fallthrough` pseudo-statements instead of `__attribute__((fallthrough))` for portability
* generate `Makefile` with clang specific warning flags
* adjust dynamic library link command

compiler:
* add target architecture define for C library source preprocessing
* select library extension based on target architecture

tools/tester/test_db.c2:
* reduce read buffer size to 128k in `Db.testFile`, causes stack overflow
* make `Db.printIssues` return error status

tools/tester/tester.c2:
* use `snprintf` instead of `sprintf`
* control verbosity with `-t` and `-v`
* control threading with `-n` and `-j[#]` (with optional number of threads)
* more flexible options parser: accept multiple options before or after file/dir
* fix crash bug on empty filename arg, do not strip `/` from root directory

### DIFF
--- a/bootstrap/.gitignore
+++ b/bootstrap/.gitignore
@@ -1,1 +1,5 @@
 c2c_bootstrap
+c2c_bootstrap2
+globals.i
+globals
+bootstrap-darwin-arm64.c
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -46,19 +46,7 @@ rebuild:
 	mv -f ../output/c2c/cgen/build.c bootstrap.c
 	$(C2C) --target arm64-apple-darwin-macho --test c2c
 	mv -f ../output/c2c/cgen/build.c bootstrap-darwin-arm64.c
-	-diff bootstrap.c bootstrap-darwin-arm64.c > bootstrap-darwin-arm64.patch
-	rm bootstrap-darwin-arm64.c
-
-bootstrap-darwin-arm64.c: bootstrap-darwin-arm64.patch bootstrap.c
-	patch -o $@ bootstrap.c $*.patch
-
-rebuild:
-	@echo "---- rebuilding bootstrap compiler ----"
-	$(C2C) --target x86_64-unknown-linux-gnu --test c2c
-	mv -f ../output/c2c/cgen/build.c bootstrap.c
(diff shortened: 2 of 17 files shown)


bvdberg:
This is a rather large change. Maybe you can extract the tools/tester/tester.c2 changes, since they dont appear to be related to arm64?

chqrlie:
> This is a rather large change. Maybe you can extract the tools/tester/tester.c2 changes, since they dont appear to be related to arm64?

Yes a separate patch would make sense, but I had to hack the tester to make it work on Darwin-arm64 because of the stack usage and the pthread structures sizes and alignment...  Without both fixes, the port cannot be tested, which is why I bundled them.

> The bootstrap.c is just generated by c2c itself. So changing it is not an option. We need to change c2c to incorporate the required changes.

I did not see any documentation describing how you generate `bootstrap.c` directly. Do you just copy the `output/c2c/cdef/build.c` file?

Modifying the compiler to make sure it generates a portable `build.c` is non trivial.  I can shall try and find an elegant approach for this, at least for the few system dependent clib features that are used bu the compiler, such as `stdio`, `errno`, `stat`, `fcntl`, `dirent`, `uname`.

Shipping multiple versions of `bootstrap.c` seems overkill and complicated as it would require the c2compiler to support cross compilation for the C output.

bvdberg:
yes bootstrap.c is indeed just a copy of output/c2c/build.c.
So for some changes (like the fallthrough), you also need to update the bootstrap version, otherwise it wont build anymore. For other changes (like the case a..b) that are not used inside the c2 compiler itself, this is not needed

bvdberg:
You probably need to rebase this since I merged your other patch..

chqrlie:
I have split this into separate PRs and will close it as obsolete.

PR #135, opened 2024-12-21
Add support for clang and older gcc compilers

chqrlie:
generator/:
* generate `fallthrough` pseudo-statements instead of `__attribute__((fallthrough))` for portability to compilers that do not support attributes or this particular one.
* generate **external.h** with compiler version detection and select `fallthrough` expansion accordingly

env.sh: define `C2_PLUGINS` as local directory to avoid interfering
  with installed c2 version

bootstrap:
* adjust warning flags for older gcc/clang versions
* add extra step in bootstrap/Makefile to generate c2c's build.c using plugins
* update bootstrap/bootstrap.c and bootstrap/external.h (unused?)

* update tests expected C code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 output/
 c2_output/
+c2_plugins/
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -1,8 +1,9 @@
 # This makefile is auto-generated, any modifications will be lost
 
 CC=gcc
-CFLAGS=-Wall -Wextra -Wno-unused -Wno-switch -Wno-char-subscripts -Wno-zero-length-bounds -Wno-format-overflow -Wno-stringop-overflow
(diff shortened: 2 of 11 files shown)


chqrlie:
@bvdberg: Hi, glad to contribute to a promising project!

This is preliminary work for the portability work:  the C2 compiler now generates C code with `fallthrough` statements that get compiled via a `#define` expanding them to the appropriate compiler extension, or nothing at all, depending on the actual compiler used locally.

I have split my other PR into multiple steps:
* this one
* a portability PR where C2 can cross compile to selected targets and create patches for the **bootstrap.c** file automatically for bootstrapping on these targets (currently only linux and macOS). This one needs an updated `c2_libs` with `#if` blocks for the system specific structures, definitions and code.
* another PR with unrelated modifications to the **tester**

bvdberg:
@chqrlie The changes look fine by me! Your first merged patch!

PR #138, opened 2024-12-21
Improve tools/tester

chqrlie:
* use 128k buffer instead of 1M to avoid stack overflow
* make `Db.printIssues` return true in case anything is reported
* control verbosity with `-t` (terse) and `-v` (verbose)
* control threads with `-n` (no threads) and `-j` (use treads)
* number of threads can be speficied as optional number after `-j`
* fix code to strip off trailing slash: incorrect for `"/"` and UB for `""`


bvdberg:
yes, i like it! merged

PR #139, opened 2024-12-21
Port to Darwin 64-bit, make bootstrap multiplatform

### LINKED ISSUE #127: Bootstrapping on macOS
I tried building the project on macOS using an M2 cpu.

This does not works for multiple reasons:
* the hardware is not recognized: the CPU is reported as `arm64` and the code tests for `arm_64`.
* some of the symbols in the C library are implemented differently than expected: `stdout` is a `#define` for `__stdoutp`, `errno` uses a function named `__error` instead of `__errno_location`...
* more generally, the structures for some system calls are different (eg: `stat`, `dirent`, `utsname`...) and the constants may differ too.

I wrote a patch for `bootstrap.c`, but the C2 source code must be changed to, either with `#if` or `#ifdef` directives or preferably with a generic approach using the include files on the local system.

Do you already have a working solution for this problem or would you be interested in a generic approach as suggested?

chqrlie:
bootstrap:
* keep intermediary target **c2c_bootstrap2**
* add **globals.c** to help port C library constants and structures
* add patch file for Darwin arm64 target
* add rebuild **Makefile** target to update **bootstrap.c** and patch files

common/target_info.c2:
* use `"arm64"` instead of `"arm_64"` in `Info.arch_names`
* add `Info.system_define`, `arch_define`, and `target_define` for C library source preprocessing.

generator/:
* adjust dynamic library link command
* remove `darwinStdio` hack

compiler:
* add target architecture defines for C library source preprocessing:
 - `SYSTEM_DARWIN` / `SYSTEM_LINUX`...
 - `ARCH_ARM64` / `ARCH_X86_64`...
 - `TARGET_DARWIN_ARM64` / `TARGET_LINUX_X86_64`...
* select library file extension based on target architecture
* add `char* target_triple` in `Option struct`
* add `--target` option to override recipe target configuration

* update bootstrap/bootstrap.c

closes #127

### DIFF
--- a/bootstrap/.gitignore
+++ b/bootstrap/.gitignore
@@ -1,1 +1,5 @@
 c2c_bootstrap
+c2c_bootstrap2
+globals.i
+globals
+bootstrap-darwin-arm64.c
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -4,22 +4,52 @@ CC=gcc
 CFLAGS=-Wall -Wextra -Wno-unused -Wno-switch -Wno-char-subscripts
 CFLAGS+=-Wno-unused-parameter -Wno-missing-field-initializers -Wno-format-zero-length
 CFLAGS+=-pipe -std=c99 -O0 -g
+TARGET_ARCH:=$(shell uname -s)-$(shell uname -m)
 
-all: ../output/c2c/c2c
+C2C:=../output/c2c/c2c
 
-c2c_bootstrap: bootstrap.c external.h
+ifeq (Darwin-arm64,$(TARGET_ARCH))
+BOOTSTRAP_FILE=bootstrap-darwin-arm64.c
+else
+BOOTSTRAP_FILE=bootstrap.c
+endif
+
+all: $(C2C)
+
+c2c_bootstrap: $(BOOTSTRAP_FILE) external.h
 		@echo "---- compiling bootstrap compiler ----"
-		@$(CC) $(CFLAGS) bootstrap.c -o c2c_bootstrap -ldl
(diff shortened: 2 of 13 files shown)


chqrlie:
For this version of C2 to support Darwin targets, the C2 libraries must be updated to include `#if` sections for Darwin specific definitions.  I have posted a PR for **c2lang/c2_libs**.

bvdberg:
I put some comments in the c2_libs first. I think it will be easier if that is merged first and then I'll make it a submodule in c2compiler.

Some comments:
- The way c2c handles cross-compilation is the *build.yaml* file. All the options for target triplets should be in there, no a separate cmdline argument. This has been done to avoid using 10+ arguments that normally appear when cross compiling in C. In the build.yaml you can also specify the compiler toolchain used and other options.

Getting rid of the Darwin stdio hack is very nice! I'm currently trying to wrap my head around the whole issue. Cross compilation is always complex, but with bootstrapping with bootstrap.c makes it even more so.. No matter.

I'll look at this patch further on my MacPro..

bvdberg:
I'm fixing the build the old way for Mac, also based on your patch...

bvdberg:
I've pushed a working version for Darwin x86_64. To compile for arm64, just use bootstrap_darwin_x86_64.c  and see if it works. Please also update the c2_libs submodule for that. If there are changes between darwin x86_64 and darwin arm64, please add #if ARCH_ARM64 in c2_libs

I had to fix 3 issues:
- stdin -> stdinp (still hardcoded in c2c)
- fstat/lstat/stat -> <x>64 variants (otherwise gave no error, but didn't work
- readdir -> readdir$INODE64   (hmm, but it works)

bvdberg:
There was one issue with readdir() on Darwin x86_64, but i fixed that, so now it's rebased and merged

PR #141, opened 2024-12-22
Fix undefined behavior on UTF-8 contents

chqrlie:
* cast `char` index values as `u8` to avoid indexing with negative values on invalid UTF-8 contents
* extend `Identifier_char` and `Char_lookup` arrays to 256 entries
* enable C warning `-Wchar-subscript` (no effect on c2c operation but often an indication of a programming error
* add test for invalid UTF-8 contents in identifier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,27 @@
+C2C:=output/c2c/c2c
+
+all: $(C2C)
+	@$(C2C)
+
+c2c: $(C2C)
+	@$(C2C) c2c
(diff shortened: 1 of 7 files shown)


bvdberg:
Yes i agree. Normally you dont have to commit bootstrap.c with every other change. Only when something changes that requires a new bootstrap.c. In this patch i see that it is needed because otherwise you get the char-subscript warnings.

PR #143, opened 2024-12-23
syntax: fix multiple issues with escape sequences and character constants

chqrlie:
* accept 8-bit octal espace sequences
* reject malformed hex espace sequences
* make 8-bit character literal values consistent with type `char`
* fix output of 8-bit octal and hex literals in `CharLiteral.printLiteral`
* fix formating bug in tools/c2cat.c2
* add tests for various issues


bvdberg:
merged

PR #148, opened 2025-01-19
improve number parser

chqrlie:
tester:
output standard error messages
add invalid number tests
add more conversion tests

compiler:
fix `printBinary` output for `0b0`
fix buffer lengths in `printBinary` and `printOctal` add support for hex floats
add missing float syntax
detect invalid syntax for integers and floating point constants detect values beyond 64-bit range for all integer syntaxes

### DIFF
--- a/ast/integer_literal.c2
+++ b/ast/integer_literal.c2
@@ -73,29 +73,29 @@ public fn bool IntegerLiteral.isSigned(const IntegerLiteral* e) {
 }
 
 fn void printBinary(string_buffer.Buf* out, u64 value) {
-    char[64] tmp;
-    tmp[63] = 0;
-    char* cp = &tmp[62];
-    while (value) {
-        *cp = '0' + (value & 0x1);
-        cp--;
+    char[1+(64+2-1)/2+1] tmp;
+    char* cp = &tmp[elemsof(tmp) - 1];
+    *cp = '\0';
+    for (;;) {
+        *--cp = '0' + (value & 0x1);
         value /= 2;
+        if (value == 0)
+            break;
     }
-    *cp-- = 'b';
-    *cp = '0';
+    *--cp = 'b';
+    *--cp = '0';
     out.add(cp);
 }
 
 fn void printOctal(string_buffer.Buf* out, u64 value) {
-    char[32] tmp;
-    tmp[31] = 0;
-    char* cp = &tmp[30];
+    char[1+(64+3-1)/3+1] tmp;
+    char* cp = &tmp[elemsof(tmp) - 1];
+    *cp = '\0';
     while (value) {
-        *cp = '0' + (value & 0x7);
-        cp--;
+        *--cp = '0' + (value & 0x7);
         value /= 8;
     }
-    *cp = '0';
+    *--cp = '0';
     out.add(cp);
 }
 
@@ -114,6 +114,7 @@ public fn void IntegerLiteral.printLiteral(const IntegerLiteral* e, string_buffe
         printBinary(out, e.val);
         break;
     case 8:
+        // FIXME: incorrect for negative values
         printOctal(out, e.val);
         break;
     case 10:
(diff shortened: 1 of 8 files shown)


bvdberg:
The generic flag bits in AST objects are always initialized to 0 in the base class already.

Why did you add Token.has_error? If there is a syntax error, the parser shouldn't continue,
otherwise you get a lot of other errors that obscure the first one...

chqrlie:
> The generic flag bits in AST objects are always initialized to 0 in the base class already.

Are you referring to `i.parent.parent.integerLiteralBits.is_signed = 0;` ?
For this particular case, `is_signed` should probably be passed as an argument to function `IntegerLiteral.create()`.

It is indeed a good idea to zero initialize allocated structures and avoid redundant initializers, and if this is the case for `i.parent.init(ExprKind.IntegerLiteral, loc, 1, 1, 0, ValType.RValue)`, the same behavior should apply to all `init` functions to avoid having to check individual calls.

Some initialization bugs are lurking in the code as can be demonstrated by defining `malloc` this way:
```
public fn void* malloc(c_size size) {
    void *p = calloc(size, 1);
    if (p) return string.memset(p, 0xff, size);
    return p;
}
```
I patched some missing intializers, but a generic approach seems necessary, especially for structures allocated using custom allocators.
'
> Why did you add `Token.has_error`? If there is a syntax error, the parser shouldn't continue, otherwise you get a lot of other errors that obscure the first one...

I introduced this flag to allow distinction between fatal syntax/grammar errors that generate an `Error` token and non-fatal errors that may allow further parsing and analysis.

Such errors include syntax errors on numeric literals, invalid escape sequences in string and character constants (not used yet), use of obsolete operators (`->` instead of `.`), expressions with dubious operator order...

The behavior is similar to `-Werror` that makes warnings fatal errors, but allows for further parsing and analyses before exiting the compiler with a failure status.

I initially wanted this behavior to be controlled by `--test` but the lexer does not have access to the `compiler.Options` object and passing individual options to the parser and then to the lexer as individual arguments to the `create` or `init` functions is too cumbersome.  What is the logic for not passing the options object?

bvdberg:
Literals are treated a bit differently. C2C only checks if it fits once it's used in any expression (and it should fit the type it is assigned to). This is done with all compile-time values (ctv). So at IntegerLiteral creation, it doesn't know if it's signed or not.

The approach with malloc is a nice one. I'll have to remember it.

The Lexer/parser should work the same in any case, so I dont think any cmdline option should change that. I also dont mind stopping on any Lexer error. The only disadvantage is that some tests have to be split up in multiple files.

chqrlie:
> Literals are treated a bit differently. C2C only checks if it fits once it's used in any expression (and it should fit the type it is assigned to). This is done with all compile-time values (ctv). So at IntegerLiteral creation, it doesn't know if it's signed or not.

I agree with this approach, yet the ctv should use at least one extra bit for the computations, or maintain a `negative` bit along the `u64` value to allow for proper handling of `u64.min`.

> The approach with `malloc` is a nice one. I'll have to remember it.

This could be a compilation debug option tested with a `#if` in `libs/libc/stdlib.c2i`.

> The Lexer/parser should work the same in any case, so I dont think any cmdline option should change that.

The lexer already has individual options for different behavior, passed as arguments to the `create` function. Passing a pointer to an options structure is just a convenient way to pass multiple options.

> I also dont mind stopping on any Lexer error. The only disadvantage is that some tests have to be split up in multiple files.

The parser stops at the first parsing error, but the analyser does not, which is convenient.  Getting multiple non-fatal syntax errors is also convenient, not just in the test files.

chqrlie:
> In the tokenizer you modify the number parsing to support '_'. What syntax would that be then? I also couldn't find an example in the unit tests.

Actually, I do not yet support the `_` as a digit separator, as is available on many other languages, I merely check that there is no extra characters attached to the number literal, be they letters, digits or underscores.

C23 added `'` as a digit separator, borrowed from C++ in an attempt at converging toward a common subset.  I personally resent this choice that breaks most code colorizers and much prefer `_`, which is already accepted in the syntax for *pp-numbers*.  Supporting that is not difficult but requires filtering the string before passing it to `strtoul()` or `strtod()`.

bvdberg:
ok i'll merge it. Otherwise it is a definitive improvement

PR #151, opened 2025-01-20
lexer: improve performance and error reporting

chqrlie:
* change `Action` enum ordering to make `INVALID` the default `Action` in `Char_lookup`,
* this improves invalid character detection in the lexer: no longer accept control codes as whitespace.
* this also improves `Tokenizer.lex_internal` main loop: removes a test for each token.
* check UTF-8 encoding to report invalid bytes as UTF-8 or not.
* add tests for invalid bytes

### DIFF
--- a/parser/c2_tokenizer.c2
+++ b/parser/c2_tokenizer.c2
@@ -185,6 +185,7 @@ const Keyword[] Keywords_w = {
 }
 
 type Action enum u8 {
+    INVALID = 0,  // ensure the default Action in Char_lookup is INVALID
     TABSPACE,
     IDENT_OR_KEYWORD,
     IDENT,
@@ -220,7 +221,6 @@ type Action enum u8 {
     TILDE,
     CR,
     EOF,
-    INVALID,
 }
 
 const Action[256] Char_lookup = {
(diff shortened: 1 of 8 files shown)


parser/c2_tokenizer.c2
```diff
@@ -781,7 +775,7 @@ fn void Tokenizer.lex_internal(Tokenizer* t, Token* result) {
         case CR:
             t.cur++;
             if (*t.cur != '\n') {
-                t.error(result, "unexpected char 0x%02X", *t.cur);
+                t.error(result, "unexpected char 0x%02X after CR", *t.cur & 0xFF);
```
bvdberg, line 778: Why AND with 0xFF, it can never be larger anyway since it's a char, maybe this needs to be 8F?

parser/c2_tokenizer.c2
```diff
@@ -793,7 +787,16 @@ fn void Tokenizer.lex_internal(Tokenizer* t, Token* result) {
             result.more = false;
             return;
         case INVALID:
-            t.error(result, "invalid char '%c'", *t.cur);
```
bvdberg, line 796: I like to keep the switch cases is order of the enum, so the INVALID  case should be first then

chqrlie: OK, I will move it.

bvdberg:
merged!

PR #154, opened 2025-01-21
lexer: use explicit initializers for token_names

chqrlie:
Make the token names initializers more explicitly synchronized with the enum values.


bvdberg:
nice one! (and you removed your old branches  :) )

PR #157, opened 2025-01-23
lexer: improve preprocessor

chqrlie:
* add `#ifdef`, `#ifndef`, `#elif`
* rename `#warn` as `#warning` for consistency
* fix `Tokenizer.skip_feature` to handle strings, character constants and comments correctly.
* modify `#error` and `#warning` to accept any message syntax
* report position of initial directive for missing `#endif`
* simplify enabled testing
* return all preprocessing directives in raw mode


bvdberg:
Why did you add the is_bol() function? the original code was just comparing t.cur with t.line_start, without looping/searching

chqrlie:
> Why did you add the `is_bol()` function? the original code was just comparing `t.cur` with `t.line_start`, without looping/searching

To allow for preprocessor directives to be indented, which is useful when they are nested.  The overhead is negligible and directives are rare anyway.

> in `skip_line_comment()`, you break on either `\r` or `\n`, but in case of `\r\n` you only skip the `\r`

`skip_line_comment()` should not skip the newline and I stop scanning on `'\0'`, `'\r'` and `'\n'`, which leaves the newline or EOF unread.  Stopping before the newline is necessary for `line_start` to be set correctly in the main loop.  I could swallow the `'\r'` because it should be followed by a `'\n'` but I handle it like `Tokenizer.lex_line_comment` for consistency.

> `skip_block_comment()` does not support nested blocks. I checked with gcc and they also don't support this. So I guess that's ok.

block comments do not nest in C, this is a feature.  Changing this is really not OK.  It violates the principle of least surprise, breaks programming tools, Of all popular languages that use `/* xxx */` for block comments, very few support nesting:

ActionScript, [AutoHotkey](https://en.wikipedia.org/wiki/AutoHotkey), C, C++, C#, D (nesting uses `/+ xxx +/` syntax), Go, [Java](https://en.wikipedia.org/wiki/Java_(programming_language)), [JavaScript](https://en.wikipedia.org/wiki/JavaScript), [Kotlin](https://en.wikipedia.org/wiki/Kotlin_(programming_language)), [Objective-C](https://en.wikipedia.org/wiki/Objective-C), [PHP](https://en.wikipedia.org/wiki/PHP), [PL/I](https://en.wikipedia.org/wiki/PL/I), [Prolog](https://en.wikipedia.org/wiki/Prolog), [Rexx](https://en.wikipedia.org/wiki/Rexx), [Rust](https://en.wikipedia.org/wiki/Rust_(programming_language)) (can be nested), Scala (can be nested), [SAS](https://en.wikipedia.org/wiki/SAS_language), SASS, SQL, Swift (can be nested), [V (Vlang)](https://en.wikipedia.org/wiki/V_(programming_language)), [Visual Prolog](https://en.wikipedia.org/wiki/Visual_Prolog), [CSS](https://en.wikipedia.org/wiki/CSS)

(source: https://en.wikipedia.org/wiki/Comparison_of_programming_languages_(syntax) )

Christoffer Lernö also changed this in C3, but I consider this a design error.  I suppose the goal is to allow commenting of code fragments with `/*` / `*/` but `#if 0` / `#endif` is a much more reliable and readable alternative for C programmers and prepending `//` on each line of the block is a popular feature available as a builtin command on most programmers editors.

Note as a matter of fact that commenting a block of code with `/*` / `*/` does not always work: if the sequences `/*` or `*/` appear in string literals or a line comments, the nesting will fail or cause unexpected side effects.

The c2 lexer does complain about nested block comments, flagging `/*` inside a block comment as an error, which is OK to prevent potential programmer errors, but I guess it is also OK to omit such checks in skipped blocks, just like other syntax errors (stray UTF-8 characters, malformed strings, character constants or numbers...).

PR #161, opened 2025-02-12
string_pool: improve hash table handling

chqrlie:
* make hash size configurable at create time
* Pass hash_table size to `string_pool.Pool.create`
* handle collisions using lists instead of flat arrays, this makes hashing more resilient to pathological collisions

### DIFF
--- a/ast_utils/string_pool.c2
+++ b/ast_utils/string_pool.c2
@@ -19,7 +19,21 @@ import string;
 import stdlib;
 import stdio;
 
-const u32 HASH_TABLE_SIZE = 256;
+#if 1
+/* 32-bit pools can store up to 32K unique string for a combined length of 512KB */
+type HashEntry u32;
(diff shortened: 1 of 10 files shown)


bvdberg:
I looked at the code and I'm not sure i completely understand it all. When I run the code it is faster indeed (both in parsing and analysis), so that's good. The report() function crashes, because that still uses HASH_TABLE_SIZE (this constant can be removed). In report it should be replaced with p.entry_size I guess.

Does this limit a string pool size to 256KB (13 bits)?

chqrlie:
It did limit the pool to 16K strings of up to 256KB, I just posted an update to bring these up to 32K and 512K by aligning the strings on 4 byte boundaries.
I also added an alternative using 64-bit hash entries with no such limitations.
It is faster because there are very few collisions and the string comparison is simpler.
Hash entry table reallocation is also simpler.

PR #163, opened 2025-02-16
parser: add support for preprocessor expressions

chqrlie:
* implement integer preprocessor expression parser and evaluator
  with signed 64-bit semantics (for now)
* add test file


chqrlie:
Incorrect commit message, I shall repost.

PR #164, opened 2025-02-16
compiler/analyser: use 65 bit Value for better range checking

chqrlie:
* make Value opaque
* add Value methods for all operators, initializers and conversions

### DIFF
--- a/analyser/conversion_checker_expr.c2
+++ b/analyser/conversion_checker_expr.c2
@@ -39,34 +39,16 @@ fn ExprWidth ExprWidth.mergeWider(ExprWidth* w1, ExprWidth* w2) {
     return result;
 }
 
-fn u8 val2width(u64 value) {
-    //TODO find faster way
-    if (value <= 65535) {
-        if (value <= 255) {
-            if (value <= 128) return 7;
-            else return 8;
-        }
-        if (value <= 32767) return 15;
-        else return 16;
-    }
-    if (value <= 4294967295) {
-        if (value <= 2147483647) return 31;
-        else return 32;
-    }
-    if (value <= 9223372036854775807) return 63;
-    return 64;
-}
-
 fn ExprWidth getExprWidth(const Expr* e) {
     ExprWidth result = {}
 
     if (e.isCtv()) {
         Value v = ctv_analyser.get_value(e);
         if (v.isNegative()) {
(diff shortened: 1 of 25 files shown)


bvdberg:
There seem to be more changes than just the value. I dont mind the github workflow / env / install_plugins changes, but it's better to separate them. Do you know how to do this? Otherwise I can do it.

The Value code changes do clean up a lot of duplication indeed. So no issues there :-)

What I still saw is that there are many instances of 4294967295 (and other constants) around the code. There are also defined as c2.u32_max, that might be nice to change in a later commit.

chqrlie:
> There seem to be more changes than just the value. I dont mind the github workflow / env / install_plugins changes, but it's better to separate them. Do you know how to do this? Otherwise I can do it.

I can split these changes into a separate PR and will rebase the value PR once you merge it.

> The Value code changes do clean up a lot of duplication indeed. So no issues there :-)

:)

> What I still saw is that there are many instances of 4294967295 (and other constants) around the code. There are also defined as c2.u32_max, that might be nice to change in a later commit.

Indeed.

Another question: can a module that imports `c2` define `c2` as a local variable in a function?  `c2` cannot be defined as a global variable in this case.

bvdberg:
'c2' is not a keyword (since it's more convenient during parsing to just have it as a special identifier). The use of the 'c2' module is optional. It is possible to create a variable called 'c2'. This would cause a normal conflict if also importing the c2 module of course

chqrlie:
> 'c2' is not a keyword (since it's more convenient during parsing to just have it as a special identifier). The use of the 'c2' module is optional. It is possible to create a variable called 'c2'. This would cause a normal conflict if also importing the c2 module of course

I understand this, I got a conflict writing a test case where I was importing `c2` and was using a set of global variables called `c1`,  `c2`, `c3`...

The name `c2` is very generic and the module is only used for the constants `c2.min_xxx`... and the C type aliases.  A more specific name such as `c2lang` or `limits` might be less confusing or a different approach for the limits using type properties `u32.min`...   As for the C types, a separate module `clang` would make sense as they do not really belong in `c2` :)

> It is not allowed to create a module with that name though

you mean an actual module I suppose :)  Does the pseudo-module `c2` need be imported at all? It seems to be always defined.

What about local variables?  I suppose no conflict there?

PR #165, opened 2025-02-16
parser: merge c2_keywords.c2 into token.c2

chqrlie:
* remove redundant list of keywords
* add `KWInfo` type in token.c2
* remove `c2_keywords` module


bvdberg:
It looks a lot cleaner indeed. You left the c2_keywords.c2 file in there (git delete). Also this branch is 0.5 ms slower during Analysis (not parsing) than the master branch.. i dont know why

chqrlie:
> It looks a lot cleaner indeed. You left the c2_keywords.c2 file in there (git delete).

Fixed.

> Also this branch is 0.5 ms slower during Analysis (not parsing) than the master branch.. The only reason i can see is that the order is a bit different (as, asm, assert before bool)

Reordering the keywords in the `Kind` enum is not possible, but using an array of `Kind` with the `KW_xxx` in alphabetical order for `KWInfo.init` to create them in alphabetical order is easy. I'm curious if this would change the timings.

PR #166, opened 2025-02-16
parser: add support for preprocessor expressions

chqrlie:
* implement integer preprocessor expression parser and evaluator with signed 64-bit semantics (for now)
* add test file

### DIFF
--- a/parser/c2_tokenizer.c2
+++ b/parser/c2_tokenizer.c2
@@ -159,7 +159,7 @@ const Action[256] Char_lookup = {
     ['m'] = Action.IDENT,
     ['n'] = Action.IDENT,
     ['o'] = Action.IDENT,
-// 112 -] 127
+// 112 - 127
     ['p'] = Action.IDENT,
     ['q'] = Action.IDENT,
     ['r'] = Action.IDENT,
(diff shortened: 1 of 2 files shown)


bvdberg:
Looking at the added test code, I question why we would add such functionality. The only use of these 'preprocessor' parts is feature selection. So enable/disable a feature or compare an architecture etc. Also, the recipe (and/or build file) determine which features get enabled. There are no sub-features allowed in the code (like #if feature_a   define feature_b).

What would the real life advantage be of having this expression functionality?

chqrlie:
> Looking at the added test code, I question why we would add such functionality. The only use of these 'preprocessor' parts is feature selection. So enable/disable a feature or compare an architecture etc. Also, the recipe (and/or build file) determine which features get enabled. There are no sub-features allowed in the code (like #if feature_a define feature_b).
> What would the real life advantage be of having this expression functionality?

When I was porting the libc for the BSD systems, I would have used `#if DARWIN || FREEBSD || OPENBSD` to factorize the common definitions.  I was lucky I could use `#ifdef LINUX  xxx #else common definitions #endif` but a more general syntax would definitely be useful.

If we allow some expression syntax, supporting the full gamut of C integer operators is not difficult and less surprising, allowing for `#if VERSION >= 24` or possibly something more elaborate to test OS versions.

Similarly, to select between 32 and 64 bits, we could write `#if sizeof(void*) == 4` or a something similar.  This does not work ATM and would not work with the C preprocessor, much to the dismay of C programmers, but it is self explanatory.

PR #170, opened 2025-02-22
compiler/library: improve math support

chqrlie:
* allow floating point division by zero to produce infinities and nans
* add library math functions and constants


bvdberg:
merged!

PR #171, opened 2025-02-28
compiler: bug fixes

chqrlie:
* fix `Context.alloc()`: use the last 8 bytes of the blocks
* fix `Context.alloc()` for blocks larger than c.blk_size: this was causing a crash on large initializer lists because `Context.alloc()` was unable to handle blocks larger than 16K and returned a pointer to a 16KB block.
* remove ominous `Pool.getStart()` function: was also causing a crash on large initializer lists with strings as pool was reallocated without updating `globals.names_start`
* fix missing initializer in `string_buffer.Buf.create_static`
* add test for large arrays
* disable warning `-Wstring-plus-int`
* fix c2cat compilation error


bvdberg:
Well found!

PR #172, opened 2025-03-02
compiler: improve large initializer handling

chqrlie:
* reduce expression kind to 5 bits.
* remove artificial and untested limitation on `InitListExpr` length
* use exponential growth for `ExprList` with 1.5 ratio (much faster on large initializer lists)
* add test with huge initializer list


bvdberg:
Fully agreed. The test is a big big through (200KB).. git probably compresses internally

chqrlie:
> Fully agreed. The test is a big big through (200KB).. git probably compresses internally

Yes, I hope so.  Otherwise we could generate these tests with a script, in c2 of course :)

PR #174, opened 2025-03-02
token: use Radix enum for clarity and efficiency

chqrlie:
* use `Radix` enum instead of explicit radix value `None`, `Hex`, `Octal`, `Binary`

### DIFF
--- a/ast/char_literal.c2
+++ b/ast/char_literal.c2
@@ -18,18 +18,19 @@ module ast;
 import ast_context;
 import string_buffer;
 import src_loc local;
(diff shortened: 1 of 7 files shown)


bvdberg:
Shouldn't Radix.None be Radix.Decimal? (since it indicates ten-based)

chqrlie:
`Radix.None` stands for *none specified*, which is decimal for numbers but not for character constants.

bvdberg:
Maybe change it to Default then? Since any number has a radix

PR #179, opened 2025-03-09
token: fix floating point accuracy in C backend generation

chqrlie:
* use `ftoa()` to ensure reversible floating point number conversions

### DIFF
--- a/ast/float_literal.c2
+++ b/ast/float_literal.c2
@@ -48,6 +48,7 @@ fn void FloatLiteral.print(const FloatLiteral* e, string_buffer.Buf* out, u32 in
 }
 
 public fn void FloatLiteral.printLiteral(const FloatLiteral* e, string_buffer.Buf* out) {
-    out.print("%f", e.val);
+    char[32] buf;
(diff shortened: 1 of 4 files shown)


bvdberg:
merged. I changed 2 things, removed 'default' case and by adding 'case Default'. Also remove unnecessary {} in case statement. In C2 you can do:
case 1:
   i32 a = 10;

chqrlie:
> merged. I changed 2 things, removed 'default' case and by adding 'case Default'. Also remove unnecessary {} in case statement. In C2 you can do: ```case 1: i32 a = 10;```

I got compiler errors for this or ```label: i32 a = 10;```.  Do you rely on a gcc extension to support labelled definitions?

bvdberg:
No, c2c will check if the first statement in a switch is a declaration and then wrap the body with { .. }
case 1: {
  <body>
}

bvdberg:
c2c does not generally add braces, only for case statements..

chqrlie:
> c2c does not generally add braces, only for case statements..

Of course, `c2c` should only add braces as needed.  Labelled definitions need them for the same reason case statements need them.

PR #183, opened 2025-03-12
language: let switch handle strings too.

chqrlie:
* allow strings in switch conditions
* use simple function to compute the dispatch value 0 for `nil` and 1 for `""` unlike `sswitch`, `switch(str)` does not crash for `nil` condition with non `nil`, non `default` first case
* improve `make errors` and `make warnings`
* add tests


bvdberg:
Looking good! Just some minor notes:
- i added some tests (not merged yet) that test case 10, case Str: (where Str is global str), these both give the correct error
- I tried case "a" "b" "c":     this compiles, but it seems c2c turns this into "a\142\143" (also before this patch). It does work however.  I'll also add a test case for this.
- I changed the /* "str" */  to // "str", i also removed the check 
- added a test care for max string len, the max string len (strlen) is in fact 254 bytes, updated code for this

My plan is after this patch to replace c2c's internal sswitch with this, remove the sswitch altogether and update bootstrap again.

We should also implement your TODO for duplicate case checking for this.

chqrlie:
> Looking good! Just some minor notes:
> 
> * i added some tests (not merged yet) that test case 10, case Str: (where Str is global str), these both give the correct error

Using global constant strings as case values is not supported yet and tricky to implement.

> * I tried case "a" "b" "c":     this compiles, but it seems c2c turns this into "a\142\143" (also before this patch). It does work however.  I'll also add a test case for this.

`"b"` and `"c"` get converted to octal because they start with a hex digit and we must prevent inadvertent side effects of string concatenation, such as `"\x01" "b"` that would become `"\x01b"`, interpreted by the C compiler as a single byte string `"\x1b"`.

> * I changed the /* "str" */  to // "str", i also removed the check

I was not sure if the C generator would generate code after this comment in case there is no newline after the `:`.

> * added a test care for max string len, the max string len (strlen) is in fact 254 bytes, updated code for this

Why should it be 254 bytes?  There is no null terminator between the concatenated strings. A length byte of 255 can be followed by 255 characters for the string match.

> My plan is after this patch to replace c2c's internal sswitch with this, remove the sswitch altogether and update bootstrap again.

Yes, good simplification.

> We should also implement your TODO for duplicate case checking for this.

I have code almost ready for this: I am converting the strings to their internal representation and use the string pool to provide unique ids to use the same duplication test as for integer values, using `0` for `nil`.  This also makes it easy to test for embedded null bytes (scanning for `\0` does not work in the source format as there are multiple possible representations).

bvdberg:
We will have to see whether it's 254/255 then, i tried in a
testcase (test/stmt/switch/switch_str_case_too_long.c2) and 255 bytes didn't
work. The len() in StringLiteral is a bit different in that it includes the
0 terminator.

escaping:
But both strings are just strings, so "\123" will remain a 4+1 byte strings
internally. So when comparing, it will just be
compared char by char right?

I think that if you concatenate strings, they should really be
concatenated, so any interpretation will be done of the final
result. I think "a" "b" "c" should just become "abc"

PR #185, opened 2025-03-18
common: improve process_utils.find_bin()

chqrlie:
* no longer use a local result buffer
* no longer allocate memory
* prevent potential buffer overflows
* fix `s = p + 1`, undefined behavior for `p == nil`
* handle empty PATH entries as current directory (like bash)


bvdberg:
printf magic.. The code is a bit cryptic, but it does work and indeed has less memory management. I'll merge this

PR #188, opened 2025-03-22
common: improve string utilities

chqrlie:
* add missing C library string functions
* document libc/string functions
* use `string_utils.endsWith()` instead of local copy in tester
* improve `test_utils.find()`


bvdberg:
I'm not a real fan of using /* */ comments for single lines. I think // is much easier on the eyes, dont you think?

bvdberg:
merged

PR #189, opened 2025-03-22
compiler: use converted strings from tokenizer

chqrlie:
* add `Token.name_idx` for clarity
* use `Token.text_idx` only for strings and comments
* `Token.text_len` no longer counts the null terminator, as the name implies
* simplify multi string concatenation.
* accept unicode escape sequences in strings and character constants
* report invalid UTF-8 sequences in strings and character constants
* add `StringLiteral.getTextIndex()`
* add `string_buffer.Buf.encodeBytes()` to encode string fragments with
* add `string_buffer.Buf.add_utf8()` to encode Unicode codepoints in UTF-8 backslash escapes.
* add **utf8** module for encoding and decoding Unicode code points.
* allow up to 255 bytes in string `case` labels
* detect and report duplicate string case labels
* detect and report string case labels `case` with embedded nulls bytes

### DIFF
--- a/analyser/module_analyser_switch.c2
+++ b/analyser/module_analyser_switch.c2
@@ -21,6 +21,7 @@ import init_checker;
 import src_loc local;
 import scope;
 import string_buffer;
+import string;
 
 fn void Analyser.analyseSwitchStmt(Analyser* ma, Stmt* s) {
     SwitchStmt* sw = cast<SwitchStmt*>(s);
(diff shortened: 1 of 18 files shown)


bvdberg:
A big one... I checked it and it looks great! Even the plugins modified, nice. It is indeed better to store the decoded strings and to re-encode them again when printing. It lowers the size of the main string-pool from 98.000 -> 86600 bytes somehow. well done!

PR #190, opened 2025-03-23
process_utils: improve consistency of command runner

chqrlie:
* remove `string_utils.split_cmd_args()`
* pass output buffer size to `process_utils.run2()`
* pass arguments before output filename in `process_utils.run_args()`
* factorize code between `process_utils.run()` and `process_utils.run_args()`
* improve consistency between `process_utils.run_args()` and `process_utils.run2()`
* merge common/process_utils2.c2 into common/process_utils.c2
* simplify `run_cmd` in shell_cmd_plugin
* pass target specific `make` name in `qbe_generator.build()`


bvdberg:
The patch is a bit hard to read, because you moved some functions around. Also it introduces a few issues:
error[numread] = 0  is now always called (it was prefixed with else), so it could be called with numread=-1.

This code is very tricky to test. This version was used a year in production code without issues. Of course some checks could be added, but i'm not a fan of just changing everything here, since it is very easy to introduce bugs..
Maybe you could make the patch smaller? Only changed a few things?

chqrlie:
> The patch is a bit hard to read, because you moved some functions around. Also it introduces a few issues: `error[numread] = 0` is now always called (it was prefixed with else), so it could be called with` numread=-1`.

The `else` I removed were redundant: 
```
    if (numread < 0) {
        fprintf(stderr, "Error reading pipe\n");
        return -1;
    }
    else error[numread] = 0;  // redundant else
```

> This code is very tricky to test. This version was used a year in production code without issues. Of course some checks could be added, but i'm not a fan of just changing everything here, since it is very easy to introduce bugs.. Maybe you could make the patch smaller? Only changed a few things?

Yes, the original code is a tricky and a tad obscure, with small inconsistencies between versions and potential undefined behavior for long argument strings or command lines.

I also noticed a handle leak on `output_pipe[0]` so the compiler might run out of handles on some systems for large projects.  Multiple handles are leaked in case of error.

The code for `Db.testFile()` is more readable and simpler, it closes the unnecessary handles before `exec` so it does not need the `fcntl` call for `FD_CLOEXEC`.

I am going to rework the patch to minimize diff

bvdberg:
merged

PR #191, opened 2025-03-23
compiler: improve string_utils and diagnostics output

chqrlie:
string_buffer:
* remove `string_buffer.Buf.add_line()`
* add `string_buffer.Buf.vprintf()`

string_utils:
* remove `string_utils.count_tabs()`
* add `string_utils.toLower()`
* fix UB on empty string in `stripNewLine()`
* handle `\r\n` in `stripNewLine()`
* return destination pointer in `toLower()`, `toUpper()`,  `stripNewline()`

utf8:
* add `utf8_sync()` to find character boundaries

parser_utils:
* return offset past the end in `getTokenEnd()`

diagnostics:
* improve range handling
* display line fragment for long lines
* detect and ignore region spanning multiple lines
* handle tabs anywhere on source line

### DIFF
--- a/ast/symbol_table.c2
+++ b/ast/symbol_table.c2
@@ -18,9 +18,9 @@ module ast;
 import string_buffer;
 import color;
 
+import stdio local;
 import stdlib;
 import string;
-import stdio;
 
 public type SymbolTable struct {
     u32 num_symbols;
@@ -121,8 +121,7 @@ public fn void SymbolTable.dump(const SymbolTable* t) {
         const char* name = ast.idx2name(name_idx);
         out.print("  [%2d]  %6d  %s\n", i, name_idx, name);
     }
-    out.stripNewline();  // puts will add another
-    stdio.puts(out.data());
+    fputs(out.data(), stdout);
     out.free();
 }
 
--- a/ast_utils/string_buffer.c2
+++ b/ast_utils/string_buffer.c2
@@ -109,20 +109,6 @@ public fn void Buf.add2(Buf* buf, const char* text, u32 len) {
     buf.data_[buf.size_] = '\0';
 }
 
-public fn void Buf.add_line(Buf* buf, const char* text) {
-    assert(text);
-    const char* end = text;
-    // TODO: take maximum length as argument
-    // TODO: avoid splitting in the middle of a UTF-8 encoding
-    while (*end) {
-        if (*end == '\n' || *end == '\r') break;
-        end++;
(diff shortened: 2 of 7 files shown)


bvdberg:
It took some time to compare the diagnostics.c2. I understand what you did. Mostly pulling the end loc stuff in a smaller scope. The code for skipping columns in long lines seems weird. loc_col is modified (loc_col -= skip) but never used after that.

I think i counted the tabs, to get the '^' at the right place as well, since the tabsize may differ per system, i'm not sure it still  works currently..

chqrlie:
> It took some time to compare the diagnostics.c2. I understand what you did. Mostly pulling the end loc stuff in a smaller scope. The code for skipping columns in long lines seems weird. loc_col is modified (loc_col -= skip) but never used after that.

OOPS! that's a bug, `if (col == loc.column) c = '^';` must be changed to `if (col == loc_col) c = '^';`

I added more comments to explain the methodology.

> I think i counted the tabs, to get the '^' at the right place as well, since the tabsize may differ per system, i'm not sure it still works currently..

Yes, you were counting the initial TABs for this purpose, but that is not sufficient if the source line contains TABs in the middle of the line, for example to align columns in an array initializer.  I changed the code to output a TAB in every position where there is a TAB in the source line above (or the visible portion thereof).  This does not account for TABs inside the range, but this should be rare.  I personally never use TABs (except in Makefile) but rejecting them seems unwarranted.

Another source of potential misalignment is unicode characters that may have a zero width or a double width. Such characters could be present in the source line inside strings or comments. I wrote code to support that in **qemacs**, but it is a low priority.

bvdberg:
I found 2 issues:
- one is utf8.utf8_sync, this is an anti-pattern, just name the function utf8_sync -> sync in module utf8 (no it wont clash ;) )
- the second is that the ^ is now broken, check:
const char* str = "abc"

It gives an error:
dd.c2:3:24: error: expected ';'
const char* str = "abc"
                           <- empty line, missing ^

On the master branch it gives:

dd.c2:3:24: error: expected ';'
const char* str = "abc"
                                        ^     (it should point to 1 column to the right of ")

I'll see if i can spot the issue, it's not a long line, so the code should basically do the same as before

chqrlie:
I fixed both issues.  renamed `utf8_sync` as `sync` in module `utf8`... importing it inline is not recommended.

The ^ issue was an off by one check when the error is exactly at the end of a line (with ot without a newline).

PR #195, opened 2025-03-26
compiler: extend command options

chqrlie:
* rename `parse_opts` and `parse_long_opts` arguments for clarity
* add `string_list.List` methods: `addStr()` and `del()`
* made `Recipe.addTarget()` public
* compiler: accept multiple targets or filenames on command line

### DIFF
--- a/common/string_list.c2
+++ b/common/string_list.c2
@@ -20,13 +20,13 @@ import string local;
 import stdlib;
 
 public type List struct {
-    const string_pool.Pool* pool;
+    string_pool.Pool* pool;
     u32* indexes;   // into pool
     u32 count;
     u32 capacity;
 }
 
-public fn void List.init(List* l, const string_pool.Pool* pool) {
+public fn void List.init(List* l, string_pool.Pool* pool) {
     memset(l, 0, sizeof(List));
     l.pool = pool;
     l.resize(16);
@@ -55,6 +55,23 @@ public fn void List.add(List* l, u32 name_idx) {
 
     l.indexes[l.count] = name_idx;
     l.count++;
+}
+
+// remove an index from the string list
+// return number of items removed.
+public fn u32 List.del(List* l, u32 idx) {
+    u32 j = 0;
+    u32 count = l.count;
+    for (u32 i = 0; i < count; i++) {
+        if (l.indexes[i] != idx)
+            l.indexes[j++] = l.indexes[i];
+    }
+    l.count = j;
+    return count - j;
+}
+
+public fn void List.addStr(List* l, const char* str) {
+    l.add(l.pool.addStr(str, true));
 }
 
 public fn bool List.contains(const List* l, const char* name) {
--- a/compiler/c2recipe.c2
+++ b/compiler/c2recipe.c2
@@ -61,7 +61,7 @@ fn void Recipe.addPlugin(Recipe* r, u32 name, u32 options, SrcLoc loc) {
     r.plugins.add(name, options, loc);
 }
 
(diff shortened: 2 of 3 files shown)


bvdberg:
I dont mind the renaming of comp_opts and alowing multiple targets/files, but i do mind the needless shuffling of code at the start of main(). There is no need for that.

chqrlie:
> I dont mind the renaming of comp_opts and alowing multiple targets/files, but i do mind the needless shuffling of code at the start of main(). There is no need for that.

I did not do this without purpose: I had to move the parsing past the creation of the `auxPool` so the file string lists can be allocated froim this pool, and I kept the other object definitions together...

bvdberg:
I looked at the patch again, and I really dont like the multiple files support. It's just a step in exactly the wrong direction in my opinion. Then next step is naming the output file and enabling/disabling all the warnings through cmdline options. This is exactly what i wanted to get away from.

chqrlie:
> I looked at the patch again, and I really dont like the multiple files support. It's just a step in exactly the wrong direction in my opinion. Then next step is naming the output file and enabling/disabling all the warnings through cmdline options. This is exactly what i wanted to get away from.

I guess we cannot agree on everything.  I would indeed like to be able to specify most options directly from the command line, which I find much easier for simple projects, or individual tests as happens in the **tester**.  This approach can coexist with the recipe centric approach, which has its virtues for larger projects.  Two sides of the same coin IMHO.

I would also like to transpile C2 files to C with a simple invokation, eg: using the `-c` option and possibly a `-o` option to specify the target file.  This makes allows for linking c2 modules in a project that uses other languages and vice versa.

bvdberg:
With this patch, the -f flag becomes unused. It can be removed then. Call options would then be:
c2c
c2c <target>
c2c <files>   -> automatic dummy target name?
c2c <target> <files>

What i can type is: c2c c2c test.c2.
This will overwrite the c2c target in the recipe, which is not very nice. Maybe we can check that if file(s) are specified,  the target may not already exist in the recipe. There may be an output/<target> directory.

chqrlie:
> With this patch, the -f flag becomes unused. It can be removed then. 

Yes, `-f` is implicit if one or more filenames are specified with a **.c2** extension.

> What I can type is: `c2c c2c test.c2`. This will overwrite the c2c target in the recipe, which is not very nice. Maybe we can check that if file(s) are specified, the target may not already exist in the recipe. There may be an output/target/ directory.

I agree: mixing recipe targets and filenames should probably be considered an error.

I see multiple 3 separate invocation styles:

* targetless: `c2c [help options]` for `--help`, `--version` and similar options that do not produce an actual target
* make style: `c2c [options] [recipe-target [recipe-target ...]]` one or more target names from the recipe.txt file (options can be specified to select another recipe file and override some settings)
* cc style: `c2c [options] source-file.c2 [other source files ...]` to produce a single target from a specified set of source files. Options can be specified to specify compilation settings, output directory and/or target file, the `recipe.txt` should be ignored or only used for global settings.

c2c should be extended to allow for multiple targets from the `recipe.txt`, in the order specified, just like `make`.  Without any specified target, all targets from the recipe.txt file are made.  This behaviour is slightly different from that of `make` that makes the first target and its dependents.  We could have a `default` section in the `recipe.txt` with a list of dependents to get this behavior.  Target sections could be specified for sets of related targets, eg: `c2c plugins`.

We could also have 2 separate executables for these 2 styles: `c2` for the make style and `c2c` for the cc style.

bvdberg:
merged. I have removed the -f option in an extra commit, since it was not needed anymore

PR #197, opened 2025-04-02
analyser: improve case termination test

chqrlie:
* detect more termination cases in case and default clauses.
* simplify `isTerminatingStmt()` and make it recursive.
* fix memory leaks
* add test cases


bvdberg:
Indeed some memory leaks. The only minor issues is that the source location is not completely correct. It would be nicer to put it before the next loc. That can wait however. Merged

PR #198, opened 2025-04-02
compiler: improve label grammar correctness

chqrlie:
* add `Stmt` pointer in `Label` structure.
* fix grammar for labelled statements and definitions. eg: `if (cond) done: return 0;` is now a single statement.
* accept labelled definitions (generate `label:;`)
* accept labels at end of block (generate `label:;`)
* add tests

### DIFF
--- a/analyser/module_analyser_stmt.c2
+++ b/analyser/module_analyser_stmt.c2
@@ -41,7 +41,10 @@ fn bool hasReturn(const Stmt* s) {
             break;
         case Label:
             // TODO
-            break;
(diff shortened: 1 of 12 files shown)


bvdberg:
I think having a Decl after a label is a compiler issue. The programmer should not have to type: label: ; to fix this..

chqrlie:
> I think having a Decl after a label is a compiler issue. The programmer should not have to type: label: ; to fix this..

I agree completely.  The programmer does not add the `;`, the compiler does when it generates the c code.

PR #199, opened 2025-04-05
c2cat: bug fix, cosmetic improvement

chqrlie:
* add missing initializer call (should have a warning for this)
* move object definition closer to initializer call


chqrlie:
merged

PR #206, opened 2025-04-05
file_utils: improve Reader and Writer consistency

chqrlie:
* open files in binary mode for portability
* hide `errno` member
* add error codes for partial read and write
* add `Reader.getError()` to get error string
* make `Writer.getError()` return error string instead of full message
* improve error detection
* free `Reader` data on errors
* always close file handle on errors
* improve consistency with string_buffer:
  * rename `Reader.data()` as `Reader.udata()`
  * rename `Reader.char_data()` as `Reader.data()`
* `Writer.write()` takes a `const void*` to avoid casts
* simplify `File.addCheckPoint()` error handling
* use `snprintf` to concatenate paths

### DIFF
--- a/common/file/reader.c2
+++ b/common/file/reader.c2
@@ -20,76 +20,108 @@ import c_errno local;
 import sys_stat local;
 import unistd local;
 import stdlib;
+import string;
 
 u8 empty;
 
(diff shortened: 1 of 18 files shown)


chqrlie:
Merged

PR #207, opened 2025-04-06
compiler: cosmetic changes

chqrlie:
* fix `Writer.getError()` messages
* import string as local for consistency
* add `startsWith(const char* text, const char* prefix, const char** pp)`
* more cosmetic changes


bvdberg:
Why did you add the startsWith() function? It is not used at all. We'r not building libc ;)

PR #214, opened 2025-04-13
parser: accept generalized types and expressions in sizeof

chqrlie:
Use case of identifier to resolve ambiguity at parse time for `sizeof(a[x])`:
* accept `sizeof(Name[10])` as size of array of 10 objects of type `Name`
* and `sizeof(name[10])` as size of array element `name[10]`
* if `Name` is a global constant array, this will generate an error as `Name` is not a type, but this case is unlikely to occur in regular code.


bvdberg:
merged!

PR #215, opened 2025-04-13
parser: use ... for case ranges

chqrlie:


bvdberg:
Very clean patch! It's good that c2c doesn't use this feature internally, so we dont have to update the bootstrap.

PR #216, opened 2025-04-14
parser: change Tokenizer.lookahead() API to simplify tests

chqrlie:
- make `Tokenizer.lookahead()` return the token kind and read the token value if destination is not `nil`.


bvdberg:
Simpler = better. Merged.

PR #217, opened 2025-04-14
compiler: remove redundant casts

chqrlie:


bvdberg:
Nice! They come from a time when the c2c analyser was not so great yet

PR #218, opened 2025-04-15
parser: accept generalized types and expressions in sizeof

### LINKED ISSUE #210: the sizeof operator should accept any expression
`sizeof` currently only accepts a restricted set of arguments: builtin types, identifiers and member expressions (referring to types or variables).  Allowing more general expressions is necessary to implement safer allocation and copy idioms:
```
    MyType *p;
    [...]
    p = malloc(length, sizeof(*p));
```
similarly:
```
    string.memset(array, 0, length * sizeof(*array));
```

chqrlie:
Use case of identifier to resolve ambiguity at parse time for `sizeof(a[x])`:
* accept `sizeof(Name[10])` as size of array of 10 objects of type `Name`
* and `sizeof(name[10])` as size of array element `name[10]`
* if `Name` is a global constant array, this will generate an error as `Name` is not a type, but this case is unlikely to occur in regular code.

fixes #210


bvdberg:
Based on the latest master, the unit test are broken:

test/parser/sizeof_expr_ok.c2:64: unexpected error: unknown module: 's'
test/parser/sizeof_expr_ok.c2:69: unexpected error: 'aa[0]' is not a type

chqrlie:
> Based on the latest master, the unit test are broken:
> 
> `test/parser/sizeof_expr_ok.c2:64: unexpected error: unknown module: 's' `
> `test/parser/sizeof_expr_ok.c2:69: unexpected error: 'aa[0]' is not a type`

Are you sure? I build with no errors on my system but the test file does produce errors with the current master, as expected.

bvdberg:
yes i tried everything again. Still the same errors...  I'll try on my mac later today

chqrlie:
Indeed different behavior on linux and macOS.  The CI test fails but is not reported as failure either:
```
Run . env.sh
setting C2 development environment
test/parser/sizeof_expr_ok.c2 
test/parser/sizeof_expr_ok.c2:[6](https://github.com/c2lang/c2compiler/actions/runs/15007841090/job/42170721587?pr=218#step:4:7)4: unexpected error: unknown module: 's'
test/parser/sizeof_expr_ok.c2:69: unexpected error: 'aa[0]' is not a type
RESULTS: 1101 tests, 4 threads (10[8](https://github.com/c2lang/c2compiler/actions/runs/15007841090/job/42170721587?pr=218#step:4:9)6 ok, 1 failed, 14 skipped) ran in 1324 ms

Failed test summary:
test/parser/sizeof_expr_ok.c2
```
The tester exits with a zero status after reporting the test failures... This is a bug.

Ill try and investigate why the behavior differs on macOS.

PR #219, opened 2025-04-15
compiler: improve assert statement

chqrlie:
* simplify assert statement generation, avoid function call
* use separate strings for filename and function name and assertion
* improve expression `printLiteral`, implement `printLiteral` for `sizeof`, `elemsof`, `enumMin`, `enumMax`, `offsetof`, `to_container` function calls and types with array sizes
* encode assert string (fix bugs on embedded double quotes)
* simplify `CharLiteral.printLiteral`, use `Buf.encodeBytes`

### DIFF
--- a/analyser/module_analyser_stmt.c2
+++ b/analyser/module_analyser_stmt.c2
@@ -355,8 +355,6 @@ fn void Analyser.analyseAssertStmt(Analyser* ma, Stmt* s) {
     QualType qt = ma.analyseExpr(a.getInner2(), true, RHS);
     if (qt.isInvalid()) return;
 
-    if (qt.isPointer()) a.setPointer();
-
     Expr* inner = a.getInner();
     ma.checker.check(builtins[BuiltinKind.Bool], qt, a.getInner2(), inner.getLoc());
 }
--- a/ast/assert_stmt.c2
+++ b/ast/assert_stmt.c2
@@ -19,11 +19,6 @@ import ast_context;
 import src_loc local;
 import string_buffer;
 
-type AssertStmtBits struct {
-    u32 : NumStmtBits;
-    u32 is_pointer : 1;
-}
-
 public type AssertStmt struct @(opaque) {
     Stmt base;
     SrcLoc loc;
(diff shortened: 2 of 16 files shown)


bvdberg:
looks nice!
a few questions:
- assert (x) becomes: ((x) || c2_assert(..)), why not make this: if (x) c2_assert(..);  ?
- assert generation in c2i_generator.c2 is still disabled. Why is this? (or just a todo?)
- there were still some unused warnings in the build (they are now enabled in the recipe)

chqrlie:
> * assert (x) becomes: ((x) || c2_assert(..)), why not make this: if (x) c2_assert(..);  ?

I use this expression to avoid a nasty side effect on `if` statements. If `assert` generates an `if` statement, an `else` clause may pose problems:
```
if (whatever)
    assert(something);
else
   do_something_else();
```
would generate:
```
if (whatever)
    if (something) c2_assert(...);
else
   do_something_else();
```
and the `else` would bind to the wrong `if`.
Generating an expression does not have this problem.
I considered other options:
* generating `{ if (expr) c2_assert(...); }`  but I do not like the extra scope.
* generating `do if (expr) c2_assert(...);  while(0)`  but it is even worse.
* generating `c2_assert(expression, "filename", linenumber, "function_name", "expression source")` and making `c2_assert` a macro that generates code or not depending on `NDEBUG` presence.  This would make the intent clear and preserve it in generated C code even if disable-asserts is set. I want to avoid the function call if assertions are enabled so the macro would be need in both cases and generate the expression if assertions are enabled.  This might be a better option.

> * assert generation in c2i_generator.c2 is still disabled. Why is this? (or just a todo?)

Just a TODO, I am not sure what the `c2i_generator.c2` is used for.  If c2 code is generated, shouldn't we just generate an `assert` statement?

> * there were still some unused warnings in the build (they are now enabled in the recipe)
I shall take a look.

Fixed.  More simplifications in the `AssertStmt` object.

Question: Do you object to a command line argument to select `disable_asserts` ?  I would be very handy for testing.

bvdberg:
.c2i (C2 interface) files are the 'header' files that are generated for library targets, along with a c header to compile c projects using a C2 library. Asserts are not so important is those files.

You know me, I dont like cmdline arguments for those things.. :)

chqrlie:
> .c2i (C2 interface) files are the 'header' files that are generated for library targets, along with a c header to compile c projects using a C2 library. Asserts are not so important is these files.

If asserts can be present, they should generate c2 `assert` statements, otherwise we should throw an error.

> You know me, I dont like cmdline arguments for those things.. :)

You support `-f` and `--test` for testing, and I agree a recipe file would be overkill for this.  By the same logic, you should support command line arguments for the options passed to c2c for the test purpose.  You currently create a dummy `recipe.txt` file in the target directory... accepting command line arguments to select the proper warnings and compiler options (backend, assert-disable...) is much simpler and more flexible.

For the warnings, using pragmas like gcc and clang would be clear, simple, easy to implement and usable by the programmer, unlike the comment magic only available in the tester.

bvdberg:
Yeah the -f and --test should be removed in the future maybe. If you want
to add a few options that you think are handy, please do.
I should not be too strict :)

PR #224, opened 2025-04-18
compiler: improve Expr.getStartLoc() / Expr.getEndLoc()

chqrlie:
* add `len` field for literals and parenthesized expressions
* add `len` for call expressions, loc is position of `(`.
* implement `getEndLoc()` in all expression classes
* removed parser/parser_utils.c2
* remove `getTokenEnd()`, `GetTokenEndFn`, `diags.get_token_end`, `get_token_source()`
* simplify `prev_loc` usage


analyser/module_analyser_call.c2
```diff
@@ -183,7 +184,8 @@ fn QualType Analyser.analyseCallExpr(Analyser* ma, Expr** e_ptr) {
 
     u32 expected_args = func_num_args - num_auto_args - isTypeFuncCall;
     if (call_num_args < expected_args) {
-        ma.error(call.getEndLoc(), DiagTooFewArgs, fd.getDiagKind(), expected_args, call_num_args);
+        // error loc is closing parenthesis
+        ma.error(call.getEndLoc() - 1, DiagTooFewArgs, fd.getDiagKind(), expected_args, call_num_args);
```
bvdberg, line 188: The diag was exactly where it was supposed to be, at the ')'. Now it's pointing at the last char of the last arg

chqrlie: The semantics of `endLoc()` have changed: it is no longer the starting location of the end token (the ')'), it is now always the offset of the character **after** the token.  The same holds for `endLoc` local variables, I always store the location after a given token.  For consistency, `prev_loc` should be renamed `prev_endLoc`.

This simplifies the range calculations: `{ e.getStartLoc(), e.getEndLoc() }` is the exact source fragment for a given expression, and `diag` no longer needs `getTokenEnd()` to compute the range.

For the missing arguments, we indeed want to point to the `)`, hence `call.getEndLoc() - 1` as we can assume that the function call ends with a `)`.

bvdberg:
This was a big patch to review....

I was worried that it slowed dont the parsing speed, since we now do more things we only did during actual errors before. That's usually not good, but here it does simplify things a lot (it does get rid of parser_utils.c2)

The end-loc is used in 2 situations:
- an error that points to that location
- when we give a diagnostic with a range
I was triggered by the getEndLoc() -1  in the analyser and endloc = p.tok.loc + 1 in the parser. It seems the endLoc is now specified as the byte *after* the token. For a range you would want to print: for (i=startloc, i<endloc; i++).  So exclusive of end location. Wouldn't it be more correct to define endloc as last byte of the expr? Then most of the +1/-1 could go i think). Only the Diagnostics, printRange should use '<= end' instead of '< end'.

small points:
- in exprBits, the name len is a bit vague, maybe use 'src_len', like you already use elsewhere? And then use that everywhere throughout the code. (src_loc and src_len, nice)
- in exprBits there is a new field 'len', but this is not counted in NumExprBits. This breaks the mechanism. Every sub-expr has ExprBits  as first field in its derived SubExprBits struct. I think the len field should be removed from the main ExprBits and only put in sub-expr when needed. Now all expr appear to have a bits.len field, but they dont actually.
- bitoffsets eg. a[16:10] should have a src_len field that covers the entire expr (up to ]). Could be changed after this patch as well, since they dont actually have it now.
- for things like string literals, the src_len is passed at creation, but strlen() is used  getEndLoc(). Also in IdentifierExpr / MemberExpr.-

chqrlie:
> This was a big patch to review....

Yes, but not as large as some more recent ones :)

> I was worried that it slowed dont the parsing speed, since we now do more things we only did during actual errors before. That's usually not good, but here it does simplify things a lot (it does get rid of parser_utils.c2)

I was worried too, but setting the token len in a place where we also set the token kind make its cache friendly.

> The end-loc is used in 2 situations:
> 
> * an error that points to that location

For this purpose (missing token), the offset of the byte _after_ the token is what we want and `prev_loc` is set accordingly with a very simple expression.

> * when we give a diagnostic with a range
>   I was triggered by the `getEndLoc() -1`  in the analyser and `endloc = p.tok.loc + 1` in the parser. It seems the `endLoc` is now specified as the byte _after_ the token. For a range you would want to print: `for (i=startloc, i<endloc; i++)`.  So exclusive of end location. Wouldn't it be more correct to define `endloc` as last byte of the expr? Then most of the +1/-1 could go I think). Only the `Diagnostics`, `printRange` should use `<= end` instead of `< end`.

Well I considered that, but making `endLoc` the offset _after_ the token allows for empty tokens as happens for example at end of file.  It is idiomatic in C for end pointers to point _after_ the last item.  The instances of `endLoc = p.tok.loc + 1;` are in places where the token is known to have a single character.  Writing `endLoc = p.tok.loc;` in these places would be very misleading.  In postfix unary operators, the length of the token is 2 and appears as such.  The `getEndLoc() -1` is a special case where we know that the last token has a width of `1` (the closing parenthese), I added a comment to clarify this. Using just `getEndLoc()` there would be misleading too and require an extra comment.

> * in `exprBits`, the name `len` is a bit vague, maybe use 'src_len', like you already use elsewhere? And then use that everywhere throughout the code. (`src_loc` and `src_len`, nice)

`src_len` would indeed be more explicit. I shall change this.  I am not sure about `src_loc` because `loc` already has type `SrcLoc` so is always a source location.

> * in `exprBits` there is a new field `len`, but this is not counted in `NumExprBits`. This breaks the mechanism. Every sub-expr has `ExprBits`  as first field in its derived `SubExprBits` struct. I think the `len` field should be removed from the main `ExprBits` and only put in sub-expr when needed. Now all expr appear to have a `bits.len` field, but they dont actually.

Yes I was a bit sloppy to avoid creating extra `xxxBits` structures.  I shall fix that.

> * bitoffsets eg. `a[16:10]` should have a `src_len` field that covers the entire expr (up to ]). Could be changed after this patch as well, since they dont actually have it now.

Actually not needed because `BitOffsetExpr` is just the `16:10` part so its range can be computed from the lhs and rhs.  The expression with the `[` and `]` is an `ArraySubscriptExpr` that has its own `len` that does include the `]`.

> * for things like string literals, the `src_len` is passed at creation, but `strlen()` is used  `getEndLoc()`. Also in `IdentifierExpr` / `MemberExpr`.

For string literals, computing the source length with `strlen` is a mistake. 
For `IdentifierExpr` we could use an `src_len` field but for `MemberExpr` we do not have individual nodes so computing the length of the identifier name is required is produces the correct length for sub parts of the chain.  
For the whole chain we could use a `len` bitfield but it is not strictly needed even if the chain spans multiple lines.  
I am not sure if there is a real benefit in having `MemberExpr` handle a variable length chain of identifiers. It adds complexity where the only special case should be `a.b`.

PR #227, opened 2025-04-19
compiler: accept full expressions in case ranges

chqrlie:
Case handling:
* add `expr_list.List` with dynamic allocation beyond initial set
* remove `identifier_expr_list.c2`
* remove `EnumTypeDecl.findConstantIdx`
* remove `IdentifierExpr.setCaseRange`, `IdentifierExpr.isCaseRange`
* fix `SwitchCase.instantiate`, did not return new expr
* simplify and generalize multi-case implementation
* update tests

Analyser:
* add range support in init_checker (`Checker.add2` and `Checker.find2`)
* simplify and generalize multi-case analysis

Generator:
* generate case ranges as both gcc and clang support them

Makefile:
* disable format warning when compiling bootstrap
* patch compilation output to fix filenames in error messages

### DIFF
--- a/analyser/conversion_checker_expr.c2
+++ b/analyser/conversion_checker_expr.c2
@@ -123,6 +123,11 @@ fn ExprWidth getExprWidth(const Expr* e) {
     case ImplicitCast:
         const ImplicitCastExpr* c = cast<ImplicitCastExpr*>(e);
         return getExprWidth(c.getInner());
+    case Range:
+        RangeExpr* b = cast<RangeExpr*>(e);
+        ExprWidth lhs = getExprWidth(b.getLHS());
+        ExprWidth rhs = getExprWidth(b.getRHS());
+        return ExprWidth.mergeWider(&lhs, &rhs);
     }
 
     e.dump();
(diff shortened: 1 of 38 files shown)


bvdberg:
In C you can do many weird things that make no sense at all. I would like to avoid giving developers options to do this.  Supporting things like 'case F3*3' is one of the before mentioned cases.  Ranges are fine, but I dont want to support syntax that is just plain weird. Same goes for sizeof. Maybe is some weird case sizeof(a.member) makes test, but otherwise the only justifiable arguments are a plain type, a variable, *var or Type*). Do you have any other cases that you ever use?

chqrlie:
The only reason I wrote `case F3*3:` was to verify that various forms of constant expressions are accepted in `case` clauses.  It is difficult to envision what expressions may make sense for a developer to use in a `case` clause, as long as these are constant integer expressions, the compiler must accept them.

Whether the switch expression can be a boolean and the case expressions can be `true` or `false` is another question.  C reluctantly accepts these but C2 does not at this point.

For `sizeof`, it is less of an issue and `sizeof(a + b)` may seem quite weird, but it does evaluate to the size of the integer type determined by the promotion rules for `+`.  Whether one make actual use of it or not does not matter, the grammar says `sizeof(x)` is defined for all expressions `x` and for all types except `void`, functions and incomplete types (opaque structures).

PR #232, opened 2025-04-21
compiler: use va_list instead of Va_list

chqrlie:
* allow types to start with lowercase in interface files
* use `va_list` for consistency with C


bvdberg:
This patch only seems to rename the Va_list to lower case (and fix the checks for that). What would be the advantage of that, besides breaking the naming scheme?

chqrlie:
> This patch only seems to rename the Va_list to lower case (and fix the checks for that). What would be the advantage of that, besides breaking the naming scheme?

As commented, consistency with C for something that is part of Standard C.  Principle of least surprise.

bvdberg:
I see you also applied the 'editing bootstrap.c' trick :).. .merged

PR #233, opened 2025-04-21
tester: simplifications

chqrlie:
* remove `test_utils.copy_string`
* add `test_utils.strstart`
* improve `Db.readUntil` to extract string arguments

### DIFF
--- a/tools/tester/test_db.c2
+++ b/tools/tester/test_db.c2
@@ -17,6 +17,7 @@ module test_db;
 
 import c_errno local;
 import stdio local;
+import stdarg local;
 import stdlib local;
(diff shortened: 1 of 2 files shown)


bvdberg:
In append, why not just limit size2 and always do the memcpy followed by setting the 0-terminator? This seems cleaner than the byte-for-byte copy

chqrlie:
> In append, why not just limit size2 and always do the memcpy followed by setting the 0-terminator? This seems cleaner than the byte-for-byte copy

Yes, it is possible, but this is the special case, truncation, for which efficiency is not as important as code size.

PR #238, opened 2025-04-24
compiler: handle conditional constant expressions

chqrlie:
* conditional expressions are constant expressions if all 3 branches are constant expressions.
* add test


bvdberg:
merged

PR #242, opened 2025-04-28
compiler: fix unary plus for pointers

chqrlie:
* accept unary plus for pointers
* factorise code for unary amp and other unary operators
* accept postfix operators for explicit cast expressions and                                                                                                    
  `to_container` expressions


bvdberg:
maybe we should add some tests for this. Now all tests work before and after the patch

chqrlie:
Yes, I have tests for some of these but they conflict with the `sizeof` branch...
The analyser should actually reject arguments that are not numeric in unary plus, minus and tilde operators.
I shall write another patch for this, with new tests

PR #244, opened 2025-05-01
compiler: fix some constant casts corner cases                                                                                                                  

chqrlie:
* accept cast as enum in constant expressions                                                                                                                   
* handle float/integer conversions casts in constant expressions

### DIFF
--- a/analyser_utils/ctv_analyser.c2
+++ b/analyser_utils/ctv_analyser.c2
@@ -150,10 +150,27 @@ public fn Value get_value(const Expr* e) {
         result = get_value(i.getInner());
         QualType qt = e.getType();
         qt = qt.getCanonicalType();
+        if (qt.isEnum())
+            qt = qt.getEnum().getImplType();
         assert(qt.isBuiltin());
         BuiltinType* bi = qt.getBuiltin();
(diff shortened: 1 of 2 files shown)


chqrlie:
C supports this: `int i = { 1 };` and I plan to use it for compound literals, thus allow `(i32){ 0 }`.
Compound literals are L-values so one can take their address if a pointer is expected, for example as a function argument.

chqrlie:
Let me summarize the justification for this PR:

* I am working on compound literals: the syntax is `(type){ initializer-list }` where `type` is any type including builtins and enums and the initializer list is exactly the same as the initializer list for object definitions.  It is relatively simple to implement but in the case of builtin types and enums the initializer list has braces which is currently not supported by c2.
* compound literals support seems desirable, restricting them to composite objects seems unnecessary.
* C accepts braced initializers for scalar objects. I agree that it seems weird and useless but supporting a feature does not mean encouraging its use.
* here is one possible use for braced initializers for scalars: for any type `MyType`, declaring `MyType a = { 0 }` defines an object that is zero initialized like a global object. This does not work in c2 if `MyType` is an alias for a builtin type or an enum type.

PR #248, opened 2025-05-03
compiler: accept c-style casts

chqrlie:
* add new type lookahead parser to analyse if a paren expression                                                                                                
  is potentially a cast or a compound literals                                                                                                                  
* extend `ExplicitCastExpr` to allow c-style variant                                                                                                            
* add `printLiteral` methods for `InitListExpr` and designated initializers                                                                                     
* add tests

### DIFF
--- a/ast/explicit_cast_expr.c2
+++ b/ast/explicit_cast_expr.c2
@@ -21,7 +21,8 @@ import src_loc local;
 
 type ExplicitCastExprBits struct {
     u32 : NumExprBits;
-    u32 src_len : 32 - NumExprBits;
+    u32 c_style : 1;
+    u32 src_len : 32 - NumExprBits - 1;
 }
 
 public type ExplicitCastExpr struct @(opaque) {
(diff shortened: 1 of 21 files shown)


bvdberg:
You are mixing features again.. The C-style cast would be ok I think. But then I would expect a lot more test cases.

For compound literals, there are some things I dont like:
```
    (Foo){1, 2}.bar(); 
    // this can just be written as: (which is MUCH more readable).
    Foo f = { 1, 2 }
    f.bar();
```

What I expected you could do was:
```
    test_foo( { 1, 2 } );   // so that it infers the type of the ILE.
    // instead of adding the c-style cast in front.
```

For Init expresssions, compound literals really bring no advantage at all, except adding clutter.

```
test_i32p((i32[1]){ 1 }, 1);    // seriously?!
```

chqrlie:
> You are mixing features again.. The C-style cast would be ok I think. But then I would expect a lot more test cases.

I updated the PR without the compound literals and many test cases.  I am not sure if we should remove the alternate cast syntax.  Some people might find it more explicit.

I would also favor `downcast<MyStruct*>(p)` which would check if p is a pointer and what it points to has the type of the first member of `MyStruct`, similar to `to_container(MyStruct, base, p)`

> For compound literals, there are some things I dont like:
> 
> ```
>     (Foo){1, 2}.bar(); 
>     // this can just be written as: (which is MUCH more readable).
>     Foo f = { 1, 2 }
>     f.bar();
> ```

Yes, `(Foo){1, 2}.bar()` is just a test for operator precedence.

> What I expected you could do was:
> 
> ```
>     test_foo( { 1, 2 } );   // so that it infers the type of the ILE.
>     // instead of adding the c-style cast in front.
> ```

Yes, it will be part of the next PR

> For Init expressions, compound literals really bring no advantage at all, except adding clutter.
> 
> ```
> test_i32p((i32[1]){ 1 }, 1);    // seriously?!
> ```

Again, just a test to check if arrays can be instantiated using this construct, whereas casts do not allow arrays.

Here is a use case for this kind of thing:  `printf("%032b", v)` could be turned into
`printf("%s", convert_b("0", 32, 0, v, (char[128]){}, 128)` to support binary conversions transparently on targets where the C library does not support it.

A similar system can be used for other unsupported formats, or for extended formats implemented as type functions:
```
    Point p = { 1, 2 };
    printf("point: %x\n", p);
```
Assuming a type function `Point.convert_s` is found, this could generate:
```
    printf("point: %s\n", p.convert_s("x", 0, 0, v, (char[128]){}, 128));
```

bvdberg:
finally merged :)

PR #250, opened 2025-05-04
compiler: improve Value handling and expression evaluator

chqrlie:
AST values:
* add `getValue()` type methods for literal expr nodes, returning a `Value`
* add `Value.castAs()` type method
* add `ValueKind.Error` for invalid result and evaluation errors
* add `Value` factories for unsigned, signed, float, boolean and error

Simplify ctv_analyser:
* remove hard coded strings in `Limits` array
* dispatch to more `getValue()` handlers

Expression evaluator:
* move generator/c/c_generator_pure_call.c2 to ast/ast_evaluator.c2
* support indirect pure functions calls including recursive

Improve C generator:
* add `Generator.need_const_expr`, used for static intializers
* add `Generator.emitConstExpr()` to force pure function calls evaluation
* simplify initializer generation
* pass Diags object for code generation errors


bvdberg:
A big change.. took me a while to parse it..
Overall I like it. Many small TODO's cleaned etc.

2 issues i see:
- generator should not have a Diags. All diagnostics are done during the analyser phase.
- This patch changes behaviour of: u8 b= 1; u8 a = b  + 200; // should not give an error..

bvdberg:
I usually keep the same as in C, an u8 = u8 + (literal); is allowed as long as the literal fits an u8.
Idem for other types

ast/function_decl.c2
```diff
@@ -43,6 +43,7 @@ type FunctionDeclBits struct {
     u32 has_return : 1;     // if it returns something, set during analysis
     u32 is_template : 1;
     u32 is_type : 1;     // part of FunctionTypeDecl
+    u32 has_body : 1;    // if function is defined in C2
```
bvdberg, line 46: Can also be an inline function in a library..

bvdberg:
2 open issues:
- c_generator still relies on Diags
- ast_evaluator should not be part of AST, but of analyser_utils (like ctv_analyser)

PR #256, opened 2025-05-11
Analyser: fix scope stack and init_checkers' array

chqrlie:
* fix some missing `ma.scope.exit()` calls
* fix missing duplicate default/default not at the end errors
* add more sanity checks in `init_checkers`, allow empty start
* allocate extra `init_check.Checker` objects if needed (eg: more than 4 nested switch statements)

### DIFF
--- a/analyser/init_checker.c2
+++ b/analyser/init_checker.c2
@@ -32,19 +32,27 @@ public type Checker struct {
     u32 max;
 }
 
+#if 0
 // Note: checker itself is not malloc'ed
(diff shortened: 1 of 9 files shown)


bvdberg:
You are correct with the scopes, however, on error, we propagate upwards and then the scope is reset completely. In the past i toyed around with using setjmp/longjmp for the analyser as well (same as in the parser). I forgot why I didn't implement that. I suspect is had something to do with errors that also have notes or something

PR #257, opened 2025-05-12
Analyser: improve vector types consistency

chqrlie:
* rename `label_vector.Vector` as `LabelVector` and import locally
* pass `Label` object to `LabelVector.add`
* import `name_vector` locally
* make `init()` method optional for `LabelVector` and `NameVector`, zero initialization OK
* fix *work around `resize()`* bug: odd init sizes caused reallocation


bvdberg:
merged

PR #263, opened 2025-05-21
Compiler: rename QualType.init as QualType.create                    …

chqrlie:
…                                                                                                                                                                                                                                                         * rename `QualType.init` as `QualType.create` for consistency as this

  is a factory, not a constructor or initializer.


bvdberg:
merged

PR #264, opened 2025-05-21
ast: transpose TypeRefHolder.fill as TypeRef.init

chqrlie:
* transpose calls to `TypeRefHolder.fill` as init calls for the destination `TypeRef` object.


bvdberg:
merged

PR #267, opened 2025-05-25
Parser: accept trailing comma in function argument lists

chqrlie:
* accept trailing comma in function argument list definition
* accept trailing comma in template function calls
* add tests


bvdberg:
merged

PR #268, opened 2025-05-25
Compiler: simplify Context allocator

chqrlie:
* use internal chaining for `Block` lists.
* allow for zero initialization of `Context`.

### DIFF
--- a/ast_utils/context.c2
+++ b/ast_utils/context.c2
@@ -22,47 +22,27 @@ import string local;
 #endif
 
 type Block struct {
-    u8* data;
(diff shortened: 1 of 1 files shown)


chqrlie:
> Alternately, the current block could be handled directly as fields of the `Context` object. I shall try that.

Can you try the latest version?

bvdberg:
Faster! I think your new version that's just using Context (and not Block) in most allocs is very nice indeed. My CPU also thinks so. 200 usec faster than master. Great job!

PR #269, opened 2025-05-25
Compiler: rename some init functions as initialize to avoid name coll…

chqrlie:
…isions

* rename `ast.init` as `ast.initialize`
* rename `attr.init` as `attr.initialize`
* no longer use `init_` as function argument names
* no longer use `region_` or `errno_` in Reader and `Writer`
* no longer use `func_`, `len_`... which no longer clash with keywords


bvdberg:
rebased+merged

PR #270, opened 2025-05-27
Compiler: add init calls

chqrlie:
* local variables can have a member call in the definition to define the object and call the initializer in a single statement. eg: `TypeRefHolder rtype.init();`, `Point p.init(x, y);`...

### DIFF
--- a/ast/var_decl.c2
+++ b/ast/var_decl.c2
@@ -40,6 +40,7 @@ type VarDeclBits struct {
     u32 kind : 3;
     u32 has_init_or_bitfield : 1;
     u32 has_local : 1; // local variables only
+    u32 has_init_call : 1; // local variables only
     u32 attr_weak : 1; // globals only
     u32 addr_used : 1;
     u32 auto_file : 1;  // for parameters only
@@ -228,6 +229,14 @@ public fn void VarDecl.setLocal(VarDecl* d, bool has_local) {
     d.base.varDeclBits.has_local = has_local;
 }
 
+public fn void VarDecl.setInitCall(VarDecl* d, bool has_init_call) {
+    d.base.varDeclBits.has_init_call = has_init_call;
+}
+
+public fn bool VarDecl.hasInitCall(VarDecl* d) {
+    return d.base.varDeclBits.has_init_call;
+}
+
 public fn void VarDecl.setAttrWeak(VarDecl* d) {
     d.base.varDeclBits.attr_weak = 1;
 }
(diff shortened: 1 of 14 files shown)


ast/var_decl.c2
```diff
@@ -228,6 +229,14 @@ public fn void VarDecl.setLocal(VarDecl* d, bool has_local) {
...
 }
 
+public fn void VarDecl.setInitCall(VarDecl* d, bool has_init_call) {
+    d.base.varDeclBits.has_init_call = has_init_call;
+}
+
+public fn bool VarDecl.hasInitCall(VarDecl* d) {
+    return d.base.varDeclBits.has_init_call;
+}
+
```
bvdberg, line 239: Could you also add printing it in the VarDecl.print() function?

bvdberg:
I tried to see if there would be any difference in assembly code generation between:

```
fn Foo get1(i32 v)  {
  Foo f.init(v);  // regular struct fn, init-call
  return f;
}

fn Foo.get2(i32 v) {
   Foo f = Foo.create(v);  // static struct fn
   return f;
}
```
In clang @x86_64 -O2, both calls get turned into:
```
   movl	%edi, %eax
   retq
```

Quite amazing.. so for this example it doesn't help in deciding which approach is better.

parser/c2_parser_stmt.c2
```diff
@@ -540,14 +544,34 @@ fn Stmt* Parser.parseDeclStmt(Parser* p, bool checkSemi, bool allowLocal) {
         p.consumeToken();
         initValue = p.parseInitValue(&need_semi, false);
         break;
+    case Dot:
+        if (p.tokenizer.lookahead(1, nil) == Kind.Identifier
+        &&  p.tokenizer.lookahead(2, nil) == Kind.LParen) {
+            if (has_local)
+                p.error("local qualified variables cannot have an init call");
+            if (!checkSemi)
+                p.error("this variable cannot have an init call");
```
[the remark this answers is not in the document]
chqrlie: Done

ast/var_decl.c2
chqrlie: Done

bvdberg:
The only caveat I can think of is that using a type-function that takes a const Foo* as argument is allowed. I'm not sure if we need to forbid this.

bvdberg:
merged

bvdberg:
We could just check if the init-call function takes a non-const Foo* and give an error otherwise

PR #274, opened 2025-05-31
Syntax: add support for raw strings

chqrlie:
* raw strings are delimited with sequences of backquotes
* raw strings end with a sequence of at least the same length as the initial sequence. Shorter sequences are part of the string.
* backslashes are not interpreted and newlines are accepted
* multi-line raw strings with at least 3 backslashes are parsed as markdown-like blocks: the rest of the line is ignored, each of the following lines must start with at least the same indentation which is not part of the string.

### DIFF
--- a/parser/c2_tokenizer.c2
+++ b/parser/c2_tokenizer.c2
@@ -45,6 +45,7 @@ type Action enum u8 {
     RSQUARE,
     NEWLINE,
     EXCLAIM,
+    BQUOTE,
     DQUOTE,
     SQUOTE,
     POUND,
@@ -142,11 +143,12 @@ const Action[256] Char_lookup = {
     ['Y'] = Action.IDENT,
     ['Z'] = Action.IDENT,
     ['['] = Action.LSQUARE,
-    ['\\'] = Action.TABSPACE, // only in skipped parts, treat as whitespace
+    ['\\'] = Action.INVALID, // invalid in code
     [']'] = Action.RSQUARE,
     ['^'] = Action.CARET,
     ['_'] = Action.IDENT,   // for parsing libs
 // 96 - 111
+    ['`'] = Action.BQUOTE,
     ['a'] = Action.IDENT,
     ['b'] = Action.IDENT,
     ['c'] = Action.IDENT,
@@ -402,6 +404,9 @@ fn void Tokenizer.lex_internal(Tokenizer* t, Token* result) {
             } else {
                 result.kind = Kind.Exclaim;
             }
+            return;
+        case BQUOTE:
+            t.lex_raw_string_literal(result);
             return;
         case DQUOTE:
             t.lex_string_literal(result);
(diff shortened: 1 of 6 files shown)


chqrlie:
> How will/should this interact with ifdefs? 

Raw strings are tokens like regular strings, multi-line raw strings are a single token. Preprocessing is not performed inside a raw string.  The only special treatment is the replacement of CR LR by a single newline character and the removal of the initial indentation.  If a programmer embeds an `#ifdef` inside a multi-line raw string, it will be in the contents of the string, or the compiler will reject it if it appears within the indentation space.

> Are multi-raw strings needed? (like multi-strings "a" "b" "c" )

I did not do anything special to support them, multi-strings are just sequences of string tokens, which can be regular strings or raw strings.  As a side effect, this allows preprocessing of long raw strings: they just need to be broken in separate strings:

```
const char[] Mylongstring = ```c
    #include <stdio.h>

    int main(```
#if ConfigMainArgs
    "int argc, char *argv[]"
#endif
    ```c
    ) {
        /* place your code here */
    }
    ```;
```

This example is silly and I did not plan this, it is just a side effect of regular design.

bvdberg:
hehe using your example uncovers an issue.. the empty line 3 gives an indentation error.. i dont think it should right?

chqrlie:
We could support a different syntax for raw strings, introduced in c++11, which might be easier to support in coding environments:

```c++
const char[] Mylongstring = R"xxx(
    #include <stdio.h>

    int main()xxx"
#if ConfigMainArgs
    "int argc, char *argv[]"
#endif
    R"xxx(
    ) {
        /* place your code here */
    }
    )xxx";
```
`xxx` can be any string, so it could be used to tell the reader (and the IDE) about the contents of the raw string. I dont like this syntax and c++ does not perform the indentation removal semantics, and as you can see in github, their colorizer does not recognise this oddity either.

chqrlie:
> hehe using your example uncovers an issue.. the empty line 3 gives an indentation error.. i dont think it should right?

No it should not, the empty line should produce a single newline, I shall fix that.

bvdberg:
I'm adding a test for the indentation error, but the msg is not very clear, maybe change to 'raw string missing indentation (need 3)' ?

chqrlie:
> C3 uses `......` for raw strings.

Yes, I forgot to mention C3, sorry :(
* Raw strings in C3 also use backquotes but with slightly different semantics: only a single backquote is used as the delimiter and embedded backquotes must be doubled.
* Javascript also uses backquote for *template strings*, which can embed newlines but also `$(expr)` substitution specifications and where backslash escapes are converted.
* I chose a different scheme, closer to the popular markdown syntax, that allows for a language tag and special handling of indentation.

analyser/module_analyser_call.c2
```diff
@@ -247,7 +247,7 @@ fn bool on_format_specifier(void* context, printf_utils.Specifier specifier, u32
         Expr* arg = args[fa.idx];
         QualType qt = arg.getType();
         qt = qt.getCanonicalType();
-        if (!qt.isInt32()) {
+        if (!qt.promotesToInt32()) {
```
bvdberg, line 250: Is this needed for the raw strings?

ast/qualtype.c2
```diff
@@ -340,11 +340,11 @@ public fn bool QualType.isUInt8(const QualType* qt) {
     return bi.isUInt8();
 }
 
-public fn bool QualType.isInt32(const QualType* qt) {
+public fn bool QualType.promotesToInt32(const QualType* qt) {
     const Type* t = qt.getTypeOrNil();
     if (t.getKind() != TypeKind.Builtin) return false;
     const BuiltinType* bi = cast<BuiltinType*>(t);
-    return bi.isInt32();
+    return bi.getKind() == BuiltinKind.Int32 || bi.isPromotableIntegerType();
```
bvdberg, line 347: idem

analyser/module_analyser_call.c2
chqrlie: not really related to rawstrings, it is a fix for `printf` argument checking.  I can split it into a separate PR.

PR #276, opened 2025-06-03
Analyser: accept types that promote to i32 as width/precision printf …

chqrlie:
…arguments


bvdberg:
merged, another improvement

PR #278, opened 2025-06-06
Compiler: fix trace bug

chqrlie:
* fix `Decl.getFullName()` to use string ring for return values.


bvdberg:
merged with pleasure

PR #281, opened 2025-06-07
Analyser: report non constant format strings as errors

chqrlie:
* format strings must be constant and known at compile time so they can be changed by the compiler according to the argument types.
* non-constant format strings cannot be checked so are likely to cause undefined behavior, thus are reported as errors.
* `printf` functions can still be called with any arguments via function pointers without the `@(printf_format()` attribute, initialized with a cast.


bvdberg:
merged

PR #283, opened 2025-06-08
Compiler: unify intiializers and assignment/function argument rvalues

chqrlie:
* `InitList` expressions can be used in assignments, function arguments and return values.
* we also get enum scoping in the same places.


bvdberg:
I lot to consider. I'm trying to come up with corner cases we can put in tests..

It would be nice to fix the TODO semi (c2_parser_expr.c2), otherwise it can remain open.

bvdberg:
Minor typo in commit msg: intiializers (makes git log nicer)

chqrlie:
> Minor typo in commit msg: intiializers (makes git log nicer)

Yes, I forgot to fix that.  Changed the commit line to: 
```none
Compiler: accept initializer lists in assignments, extend enum scoping
```

bvdberg:
I was looking through the code and in analyseBinaryOperator() there is a check if opcode == Assign. That got me thinking about the other assign (+= -=  /= etc). Those currently dont work without a prefix. It is questionable whether this would be useful for enums.   Enum += <int> would be ok, but Enum += Enum-Constant would be weird.

There is also the issue of  = { .. } ;   where the semicolon doesn't give an error (it should).

chqrlie:
> I was looking through the code and in `analyseBinaryOperator()` there is a check if `opcode == Assign`. That got me thinking about the other assign (`+=`, `-=`, `/=` etc). Those currently dont work without a prefix. It is questionable whether this would be useful for enums. Enum += would be ok, but Enum += Enum-Constant would be weird.

I agree, I cannot think of a good use case for enum scoping for these operators.
The ternary operator should support `my_enum = condition ? Enum_const1 : Enum_const2;` without a prefix.
Arrays could be defined to take an enum value as an index and `[Enum_const]` would be scoped without a prefix, but we would need a specific declaration syntax and I think it would be more useful to support the C3 enum attributes instead.

> There is also the issue of `var = { .. } ;` where the semicolon doesn't give an error (it should).

I am not sure it should give an error:  The `;` should be ignored as it is redundant and this should be implemented generically for expression statements by looking at the last token or using a flag in the `Parser` object.  There are corner cases where the `;` help readability.

> What about?: `Point p = { 1, 2 } p = { 3, 4 } = { 5, 6 }`

This should give an error on the second `=`. But `Point p1; Point p2; p1 = p2 = { 1, 2 }` should work.

> (i pushed a change to master to get a better error location for that)

I'll take a look.

bvdberg:
> > I was looking through the code and in `analyseBinaryOperator()` there is a check if `opcode == Assign`. That got me thinking about the other assign (`+=`, `-=`, `/=` etc). Those currently dont work without a prefix. It is questionable whether this would be useful for enums. Enum += would be ok, but Enum += Enum-Constant would be weird.
> 
> I agree, I cannot think of a good use case for enum scoping for these operators. The ternary operator should support `my_enum = condition ? Enum_const1 : Enum_const2;` without a prefix. Arrays could be defined to take an enum value as an index and `[Enum_const]` would be scoped without a prefix, but we would need a specific declaration syntax and I think it would be more useful to support the C3 enum attributes instead.
> 
> > There is also the issue of `var = { .. } ;` where the semicolon doesn't give an error (it should).
> 
> I am not sure it should give an error: The `;` should be ignored as it is redundant and this should be implemented generically for expression statements by looking at the last token or using a flag in the `Parser` object. There are corner cases where the `;` help readability.
> 
> > What about?: `Point p = { 1, 2 } p = { 3, 4 } = { 5, 6 }`
> 
> This should give an error on the second `=`. But `Point p1; Point p2; p1 = p2 = { 1, 2 }` should work.
It complains about missing a semi-colon after }  :-)
> 
> > (i pushed a change to master to get a better error location for that)
> 
> I'll take a look.

chqrlie:
> The Tokenizer.charAt() is really nasty, since it breaks the clean interface of tokens... I'll see if I can find a nice solution.

A cleaner solution is to store the kind of the last token in the Parser object and test that.

bvdberg:
Yes, changes in a new commit are much easier to follow for me. I think the new error message is indeed an improvement.

Language wise, i prefer to keep the variations in syntax minimal. So making semicolons optional is not good then. Either they are required or forbidden. I think allowing semicolons after all statements is ok. This does create a small difference with globals (where they are forbidden). That is the price of the init-list in expressions.

chqrlie:
> Language wise, i prefer to keep the variations in syntax minimal. So making semicolons optional is not good then. Either they are required or forbidden. I think allowing semicolons after all statements is ok. This does create a small difference with globals (where they are forbidden). That is the price of the init-list in expressions.

I agree. This would definitely simplify the parser since `need_semi` would no longer be needed.

chqrlie:
> > ok lets make it so. Shall I change it on top of your commit or will you?
> 
> I'll give it a try.

Well this is a non trivial change: requiring the `;` in statements, as opposed to `struct` definitions would require a double bootstrap.  We would first need to accept the semicolon, then bootstrap, then we can require it and add the missing ones, then bootstrap again later (not immediately required).

Let's be more precise: in which *statements* would the trailing semicolon be required?

* `struct` definitions: no
* global struct/array initializers: yes/no?
* incremental struct/array additions: yes/no?
* local struct/array initializers: yes
* struct/array assignments: yes
* returning struct/array: yes

chqrlie:
I pushed a patch to make the semicolons optional so you can bootstrap.

PR #287, opened 2025-06-10
Compiler: add min and max enum properties

chqrlie:
* add `min` and `max` properties for `enum` types.
* `MyEnum.min` is an alias for the smallest `MyEnum` enum constant, with the same implementation type.
* `MyEnum.max` is an alias for the largest `MyEnum` enum constant, with the same implementation type.
* this could replace `enum_min` and `enum_max`, whose type is not well defined.
* update tests

### DIFF
--- a/analyser/module_analyser.c2
+++ b/analyser/module_analyser.c2
@@ -75,6 +75,8 @@ public type Analyser struct @(opaque) {
 
     u32 check_idx;
     init_checker.Checker[4] init_checkers;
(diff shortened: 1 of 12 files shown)


bvdberg:
Again i'm surprised by how little code changes this feature requires. Enum.min looks better than enum_mix(Enum) indeed. I've merged this and put it on the list that require a bootstrap update..thx!

PR #289, opened 2025-06-11
Compiler: add builtin properties for min/max etc.

chqrlie:
* add support for builtin properties
* handle `min`, `max` and `width` for integer types
* handle `min`, `max`, `width`, `inf`, `nan` for floating point types
* allow builtin names as module names
* create c2 library with builtin property modules
* add c2 module in c2 library and remove `c2module_loader.c2`
  - `min_xxx` and `max_xxx` definitions should be removed after boostrap
  - C types should be removed and replaced with `extern "C" {}` blocks
* update tests


analyser/module_analyser.c2
```diff
@@ -572,7 +572,9 @@ fn void Analyser.analyseGlobalVarDecl(Analyser* ma, VarDecl* v) {
 
     d.setType(res);
 
-    ma.checkName(d, res.isConstant());
+    // TODO: move to the parser
```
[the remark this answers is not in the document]
chqrlie: The parser can easily know if the variable definition is `const` or not: it is a constant if the type starts with `const` and does not have indirections. So it could test the constraint on the name, but delaying this to the analyser is fine too. 

Actually, if you can fold the `const`ness into user types (eg: `type Foo const u32;`) then the parser cannot know.  I shall remove this comment.

ast/module.c2
```diff
@@ -70,7 +70,8 @@ public fn void Module.free(Module* m) {
 public fn void Module.setUsed(Module* m) { m.is_used = true; }
 public fn bool Module.isUsed(const Module* m) { return m.is_used; }
 
-public fn void Module.setInternal(Module* m) { m.is_internal = true; }
+// TODO use use a keyword or an attribute for this
```
bvdberg, line 73: Internal means it's in the user code and not a library. No keyword/attribute is needed since it's a compiler internal thing

libs/c2/manifest.yaml
```diff
@@ -0,0 +1,23 @@
+
+info:
+#    name: "C library"
+    language: c
+    type: library
+    kinds:
+        - static
+        - dynamic
```
bvdberg, line 8: language can be c2, kinds only static. In the future i want to add a 'source ' kind as well for libraries.

parser/c2_parser.c2
```diff
@@ -189,9 +189,11 @@ fn void Parser.error(Parser* p, const char* format @(printf_format), ...) @(nore
     longjmp(&p.jmpbuf, 1);
 }
 
-fn void Parser.parseModule(Parser* p, bool is_generated) {
+fn void Parser.parseModule(Parser* p, bool is_interface, bool is_generated) {
     p.expectAndConsume(Kind.KW_module);
-    p.expectIdentifier();
+    // accept builtin types as module names
+    if (!(p.tok.kind >= Kind.KW_bool && p.tok.kind < Kind.KW_void))
```
bvdberg, line 195: nicer to add function in token.c2 like is_keyword(), but then is_builtin_type()?

parser/c2_parser.c2
```diff
@@ -189,9 +189,11 @@ fn void Parser.error(Parser* p, const char* format @(printf_format), ...) @(nore
...
+        p.expectIdentifier();
 
     const char*modname = p.pool.idx2str(p.tok.name_idx);
```
bvdberg, line 198: p.tok.name_idx is only valid if the token.kind is Identifier.

parser/c2_parser_expr.c2
```diff
@@ -801,8 +823,11 @@ fn Expr* Parser.parseFullIdentifier(Parser* p) {
 */
 fn bool Parser.parseAsType(Parser* p, bool* has_brackets) {
     const Kind kind = p.tok.kind;
-    if (kind >= Kind.KW_bool && kind <= Kind.KW_void) return true;
-    if (p.tok.kind != Kind.Identifier) return false;
+    if (kind == Kind.KW_const || kind == Kind.KW_volatile) return true;
+    if (p.tok.kind != Kind.Identifier) {
+        if (kind >= Kind.KW_bool && kind <= Kind.KW_void)
```
bvdberg, line 828: can use same token helper function as above is_builtin_type?

ast/ast.c2
```diff
@@ -97,7 +99,7 @@ public fn void AST.setPtr(AST* a, void* ptr) { a.ptr = ptr; }
 
 public fn void* AST.getPtr(const AST* a) { return a.ptr; }
 
-public fn Module* AST.getMod(const AST* a) { return a.mod; }
+public fn Module* AST.getMod(const AST* a) @(unused) { return a.mod; }
```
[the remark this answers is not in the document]
chqrlie: That's true, but this also prevents other programs from ever using the module API from a different module.  I have mixed feelings about removing the `public` qualifier from APIs that it makes sense to publish.  Maybe `@(unused_public)` would make this more acceptable?

ast/ast.c2
```diff
@@ -117,6 +119,10 @@ public fn bool AST.isGenerated(const AST* a) @(unused) {
     return a.is_generated;
 }
 
+public fn bool AST.isInterface(const AST* a) @(unused) {
```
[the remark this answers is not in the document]
chqrlie: Same idea.

libs/c2/manifest.yaml
chqrlie: Good points, I shall remove the `dynamic`.

parser/ast_builder.c2
```diff
@@ -128,8 +129,10 @@ public fn void Builder.actOnImport(Builder* b,
 
     ImportDecl* old = b.ast.findImport(mod_name);
     if (old) {
-        b.diags.error(mod_loc, "duplicate import of module '%s'", idx2name(mod_name));
-        b.diags.note(old.asDecl().getLoc(), "previous import is here");
+        if (mod_loc) {
+            b.diags.error(mod_loc, "duplicate import of module '%s'", idx2name(mod_name));
+            b.diags.note(old.asDecl().getLoc(), "previous import is here");
+        }
         return;
```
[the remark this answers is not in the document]
chqrlie: No, but I want to be able to import modules that may already be imported without an error message (this is used for the implicit import of the builtin modules, and could also be used to import the `c2_assert`, the `c2_strswitch`, the trace code as c2 source from the library.

Instead of adding another argument to make the error optional, I use a null `SrcPos` to disable the error.

parser/c2_parser.c2
chqrlie: Good point.

parser/c2_parser.c2
chqrlie: No, it is also valid for keywords :)  No change needed, I should document this.

PR #290, opened 2025-06-11
Parser: improve error message handling

chqrlie:
* use single `on_error` handler with error level and message arguments
* remove `Warning` token type, never returned anyway.
* improve `#error` and `#warning` message parsing consistency
* make `num_error` messages non fatal
* fix `#warning` behavior

### DIFF
--- a/parser/c2_tokenizer.c2
+++ b/parser/c2_tokenizer.c2
@@ -253,7 +253,8 @@ public type Feature struct {
     bool is_else;   // inside the #else block
 }
 
-public type HandlerFn fn void (void* arg, SrcLoc loc);
+public type ErrorLevel enum u8 { Note, Warning, Error, FatalError }
+public type ErrorFn fn void (void* arg, ErrorLevel level, SrcLoc loc, const char* msg);
 
 public type Tokenizer struct {
     const char* cur;
@@ -265,9 +266,8 @@ public type Tokenizer struct {
 
     string_pool.Pool* pool; // no ownership
     string_buffer.Buf* buf; // no ownership, used for strings and character constants
-    HandlerFn on_error;
-    HandlerFn on_warning;
-    void* fn_arg;
+    ErrorFn on_error;
+    void* on_error_arg;
 
     // Feature handling
     Feature[constants.MaxFeatureDepth+1] feature_stack;
@@ -278,7 +278,7 @@ public type Tokenizer struct {
 
     char[256] error_msg;
 }
-static_assert(416, sizeof(Tokenizer));
+static_assert(408, sizeof(Tokenizer));
 
 public fn void Tokenizer.init(Tokenizer* t,
                               string_pool.Pool* pool,
@@ -287,9 +287,8 @@ public fn void Tokenizer.init(Tokenizer* t,
                               SrcLoc loc_start,
                               const keywords.Info* kwinfo,
                               const string_list.List* features,
-                              HandlerFn on_error,
-                              HandlerFn on_warning,
-                              void* fn_arg,
+                              ErrorFn on_error,
+                              void* on_error_arg,
                               bool raw_mode)
 {
     string.memset(t, 0, sizeof(Tokenizer));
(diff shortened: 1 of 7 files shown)


chqrlie:
> It seems to make parsing slower... 

This is surprising, the changes should not affect the inner loops, only the error cases.

> What was it that you were trying to improve?

I was just trying to improve the consistency:
* both warnings and errors should be produced synchronously,
* `#warning` used to output an error with `on_warning`, itself using `diag.error` instead of `diag.warn`
* `num_error` are errors, not warnings, but non fatal.
* `#error` returned an error token without showing the error message.

parser/c2_tokenizer.c2
```diff
@@ -1490,28 +1488,36 @@ fn bool Tokenizer.at_bol(Tokenizer* t) {
...
+            msg.add2(t.pool.idx2str(result.text_idx), result.text_len);
+        } else {
+            msg.add2(t.input_start + (result.loc - t.loc_start), result.len);
+        }
+        last_loc = result.loc + result.len;
     }
-    char[constants.MaxErrorMsgLen+1] msg;
-    string.memcpy(msg, start, len);
-    msg[len] = 0;
+    msg.size();  // ensure null terminator
```
bvdberg, line 1506: debug leftover?

parser/c2_tokenizer.c2
```diff
@@ -1490,28 +1488,36 @@ fn bool Tokenizer.at_bol(Tokenizer* t) {
...
     const char* start = t.cur;
-    while (*t.cur != '\0' && *t.cur != '\r' && *t.cur != '\n')
-        t.cur++;
-    usize len = cast<usize>(t.cur - start);
-    if (len > constants.MaxErrorMsgLen) {
-        t.error(result, "error msg too long (max %d bytes)", constants.MaxErrorMsgLen);
-        return true;
+
+    // parse pptokens instead of raw text
+    string_buffer.Buf* msg = string_buffer.create_static(elemsof(t.error_msg), false, t.error_msg);
```
bvdberg, line 1493: string_buffer.create_static does a malloc (surprised me as well), but old code seemed to be much simpler. If that used the  t.error_msg buffer instead of a stack...  The string buffer is now barely used (it does a memcopy internally). Also why parse to Eof instead of newline?

parser/c2_tokenizer.c2
chqrlie: More anticipation of future feature :)

String buffers currently null terminate the array for each individual call. I suggest only adding the null terminator when the pointer or the size is actually needed.  This would reduce the number of redundant writes and potentially improve the inlining of single character additions.

parser/c2_tokenizer.c2
chqrlie: > string_buffer.create_static does a malloc (surprised me as well),

The `string_buffer.Buf` object would not need to be allocated in this case, as well as many other ones, if we had separate `string_buffer.Buf.init(Buf* b, const char *buf, u32 size, bool reallocate)` function called by both `string_buffer.create` and `string_buffer.create_static`.

> but old code seemed to be much simpler. 

The `#error` and `#warning` features should use preprocessed tokens like `#if` for multiple reasons:
* for consistency with the C parser.
* to simplify language support in editors and code colorizers (simpler if we follow the already supported C semantics)
* so comments are parsed as such on these lines and thus can be used by the tester, or the programmer.

> If that used the t.error_msg buffer instead of a stack... The string buffer is now barely used (it does a `memcpy` internally).

It does a bunch of `strcpy` and `strcat` with truncation.  We could indeed write and use `strlcpy`-like functions instead.

> Also why parse to`Eof` instead of newline?

Because `t.lex_preproc(result)` returns `Eof` on both the actual end of file and the first newline encountered. `lex_preproc` is a wrapper for `lex_internal` that sets the `stop_at_eol` for this very purpose.  I shall add a comment to clarify this side-effect.

PR #293, opened 2025-06-16
Compiler: accept lists of declarations

chqrlie:
* accept commas in variable declarations and struct member definitions
* type must not be a pointer type nor an array type
* add tests

### DIFF
--- a/ast/decl_stmt.c2
+++ b/ast/decl_stmt.c2
@@ -16,13 +16,14 @@
 module ast;
 
 import ast_context;
+import constants local;
 import string_buffer;
 
 import string local;
 
 type DeclStmtBits struct {
     u32 : NumStmtBits;
-    u32 count : 5;
+    u32 count : MaxMultiDeclBits;
 }
 
 public type DeclStmt struct @(opaque) {
--- a/parser/c2_parser_stmt.c2
+++ b/parser/c2_parser_stmt.c2
@@ -17,6 +17,7 @@ module c2_parser;
 
 import stmt_list;
 import ast local;
+import constants local;
 import token local;
 import src_loc local;
 import string_list;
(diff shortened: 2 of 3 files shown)


ast/decl_stmt.c2
```diff
@@ -18,35 +18,52 @@ module ast;
...
 
 public fn DeclStmt* DeclStmt.create(ast_context.Context* c,
-                                      VarDecl* decl)
+                                    VarDecl** decl, u32 count)
 {
-    DeclStmt* s = c.alloc(sizeof(DeclStmt));
-    s.base.init(StmtKind.Decl, decl.asDecl().getLoc());
-    s.decl = decl;
+    DeclStmt* s = c.alloc(sizeof(DeclStmt) * sizeof(VarDecl*) * count);
+    s.base.init(StmtKind.Decl, decl[0].asDecl().getLoc());
```
bvdberg, line 37: The alloc seems a bit much.... a * b * c, probably a + b * c  ?

chqrlie: Good catch!  updated the PR and fixed the stats too.

parser/c2_parser_stmt.c2
```diff
@@ -513,66 +513,80 @@ fn Stmt* Parser.parseWhileStmt(Parser* p) {
     return p.builder.actOnWhileStmt(loc, cond, then);
 }
 
-fn Stmt* Parser.parseDeclStmt(Parser* p, bool checkSemi, bool allowLocal) {
+fn Stmt* Parser.parseDeclStmt(Parser* p, bool checkSemi, bool allowLocal, bool allowComma) {
+    VarDecl*[31] decls;
```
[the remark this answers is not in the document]
chqrlie: Yes, let's make it 15, the count is stored in a bit-fields in the `DeclStmtBits` object.  I prefer to store the actual count eventhough a 0 count seems useless.  I just added constants for both the number of bits and the count itself.

bvdberg:
yes nice!

PR #297, opened 2025-06-21
Analyser: force parentheses for ambiguous operator combinations

chqrlie:
* flag use of combined operators with unintuitive relative precedence as errors
* fix code like `if (buf.st_mode & S_IFMT == S_IFREG)` that now requires parentheses:  `if ((buf.st_mode & S_IFMT) == S_IFREG)`
* update tests


bvdberg:
Nice..

I dont understand the C_prec is the binary-operator analyser. The c2 only i can understand.

bvdberg:
merged...

PR #298, opened 2025-06-22
Compiler: simplify IntegerLiteral, never negative

chqrlie:
* remove `is_signed` member and `isSigned()` method in `IntegerLiteral`
* simplify output, constant evaluator and generator


bvdberg:
merged

PR #299, opened 2025-06-23
C generator: fix code generation for if stmt and multi-decl

chqrlie:
* wrap `if` statement with declaration in condition in a block
* generate single statement for multi-declarations
* add tests


bvdberg:
The modified unit tests dont seem to match the changes right? Multi var with local keyword?

chqrlie:
> The modified unit tests dont seem to match the changes right? Multi var with local keyword?

I did not change the commented line about the `local` keyword in the c2i generator, just generated a comma separated list of definitions to match the source code and fix the `for` initial clause syntax.

I guess it should just be uncommented, I don't know why it is disabled.

PR #301, opened 2025-06-23
compiler: extend cast syntax for initlists

chqrlie:
* parse compound cast expressions: depending on surrounding code, explicit casts can be more readable than implicit conversions as function arguments or return values.
* add tests


bvdberg:
We had 'compound literals' before this commit right? Only without the cast. The type was inferred from the LHS. Adding the cast only seems to make the code longer, or am I missing something?

bvdberg:
What would happen with?

`Expr e = (SubExpr){ 1, 2, 3 };`

chqrlie:
Closing this PR for now. I may reopen it if and when we get implicit base casts and structure pointer decay

PR #312, opened 2025-06-29
Compiler: Use full name for symbols from builtin modules

### LINKED ISSUE #309: u8.max causes C compilation error
```
public fn i32 main() {
    u16 max = u8.max;
    return 0;
}
```

generates:

```c
// --- module u8 ---

#define max 255


// --- module test ---

int32_t main(void);

int32_t main(void)
{
   uint16_t max = max; // error
   return 0;
}
```

chqrlie:
* builtin modules are external modules whose name is that of a builtin type                                                                                     
* use full name for const definitions too.                                                                                                                      
* add missing char module in libs/c2/manifest.yaml                                                                                                              
* add regression test                                                                                                                                           
                                                                                                                                                                
Fixes #309

### DIFF
--- a/generator/c/c_generator.c2
+++ b/generator/c/c_generator.c2
@@ -157,10 +157,13 @@ fn void Generator.emitCNameMod(Generator* /*gen*/, string_buffer.Buf* out, const
         const char* cname = d.getCName();
         if (cname) {
             out.add(cname);
-        } else {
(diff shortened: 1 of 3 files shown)


chqrlie:
I have simplified this PR: all constant definitions now use the full name in the `#define`, removed the builtin special case.  Should be ready for merging.

PR #313, opened 2025-06-29
parser: disambiguate more casts on unary operators

### LINKED ISSUE #305: Parsing of new cast sometimes goes wrong
When casting to  user type (ptr), followed by & the parser says 'expected expression'. 

```
type Inner struct {
   i32 i;
}

type Outer struct {
  Inner i;
}

fn void test1(Outer o) {
   Inner* i = (Inner*)&o.i;  // -> error
   // also for assignment 
}
```

Also instead of & , * also goes wrong:  (Type*) *ptrptr;

chqrlie:
* disambiguate casts on unary operators `+`, `-`, `*` and `&` using identifier case.

Fixes #305


bvdberg:
merged

PR #314, opened 2025-06-30
Compiler: move c2_trace functions to libc/c2trace.c2i

chqrlie:
* pass `trace_calls` option to parser to generate call to `c2_trace_init()` from `main()` and thus pull library functions.
* add c2_trace.c2i module in libs/libc
* fix `StdlibCompareFn` function type to avoid typedef issue
* fix stack overflow on recursive inline functions
* generate call tracing tables in **c2_trace_tables.c** for fast builds

Fix external array definitions:
* accept array definitions without size in interface files
* fix external array definitions in C generator
* fix test/interface/incremental_array.c2t: fix external array definition
* fix test/interface/public_by_array_expr.c2t: fix external array definition
* fix test/interface/muti_dimensional_array.c2t: `Row` and `Columns` definitions were reversed, 2D external array definition in C was incorrect


bvdberg:
This patch seems to combine 2 things:
- arrays in c2i files
- c2trace refactoring

There is a couple of things i think should change:
- c2trace is now part of libc, with has nothing to do with it
- tracing is now everywhere, in the parser, analyser and generator (it used to be only generator)
- it adds more exceptions (if-statements) for tracing

Overall I think we need some other mechanism to write 'internal' code. Writing this is a c2 file is much better,
than writing it in (raw-)strings. This code should be compiled in IMO, since I dont want the c2c binary to need
internal stuff after compilation. I'll give this some thought..

chqrlie:
I simplified the port: the source module `libs/c2/c2_trace.c2` is added to the target if tracing is enabled and the tables are generated in another module `c2_trace_tables` for which an interface file is added to the c2 library.

I had to fix the C generator for extern tables as these do not really decay to pointers. (further simplifications pending).

I shall add a way to incorporate the c2 library files in the c2c compiler once the `assert` and string switch code is moved to c2.

This PR should be ready for merging.

chqrlie:
I just rebased this PR on the latest master. It should be ready to merge now!

bvdberg:
I'm trying to find a solution where we dont need the lib/ files for tracing/assert. These should just be inside the c2c executable. Would that be possible if we added a command in the recipe to embed files?

$embed mymodule mysymbol filename.c2 <null-terminate>

mymodule could be an existing/new module. If existing it must be internal (not in a lib)
mysymbol would be the symbol name that appears in the module. Must start with a capital letter

for the symbol, we have some options:
- make it a struct type (name TBD) with a data + size field (same as the load_file plugin)
- generate const void* Symbol_data = .. + const usize Symbol_size = ..

Would that be enough to remove those files in lib/ ?

bvdberg:
I think the name could be 'embeds' perhaps?

bvdberg:
Yeah it looks good. The rebase gives some conflicts, so i'll wait for your rebase and then merge it..

chqrlie:
> Since we have now less code to parse, I was hoping it would be faster.. But there is no noticeable difference in parsing. Analysis is 6% slower through... any ideas why?

Parsing is not affected because the difference in code length is negligible (the c2trace.c2 code was not parsed before either). The @embed attribute is handled at analysis time and the embedded code is always loaded as a string (but not parsed), which adds some additional I/O and processing during analysis compared to before where the c2trace.c2 code was neither loaded nor parsed for regular builds.  The C generation time might be affected too since we now always output the code encoded as a string.

PR #315, opened 2025-07-02
Fix external array definitions:

chqrlie:
* accept array definitions without length in interface files
* reject `elemsof` on arrays of unknown length
* fix external array definitions in C generator
* fix test/interface/incremental_array.c2t: fix external array definition
* fix test/interface/public_by_array_expr.c2t: fix external array definition
* fix test/interface/muti_dimensional_array.c2t: `Row` and `Columns` definitions were reversed, 2D external array definition in C was incorrect

### DIFF
--- a/analyser/module_analyser.c2
+++ b/analyser/module_analyser.c2
@@ -563,7 +563,7 @@ fn void Analyser.analyseGlobalVarDecl(Analyser* ma, VarDecl* v) {
                 return;
             }
         } else {
-            if (!init_expr) {
+            if (!init_expr && !d.isExternal()) {
                 ma.error(d.getLoc(), "array-type variable '%s' needs an explicit size or an initializer", d.getFullName());
                 return;
             }
--- a/analyser/module_analyser_builtin.c2
+++ b/analyser/module_analyser_builtin.c2
@@ -59,6 +59,13 @@ fn QualType Analyser.analyseSizeof(Analyser* ma, BuiltinExpr* e) {
         if (qt.isInvalid()) return QualType_Invalid;
 
         // TODO extract to function
+        if (qt.isArray()) {
+            const ArrayType* at = qt.getArrayTypeOrNil();
+            if (at && !at.hasSize()) {
+                ma.error(inner.getLoc(), "sizeof cannot be used on arrays of unknown length");
+                return QualType_Invalid;
+            }
+        }
         if (qt.isStruct()) {
             StructType* st = qt.getStructType();
             const StructTypeDecl* std = st.getDecl();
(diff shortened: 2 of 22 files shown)


bvdberg:
I saw your TODO's in the unit tests. This is a hard question, keep constants in .c2i/.h files and inline them for .c files?
For the arrays, since we now have the array and not the pointer in the interface, the user can do elemsof(), so doesn't need any other number for that.

Zero sized arrays should not be in the interface, so we should indeed add the size explicitly during generation.

Did you test that generating a library with an array in the interface and then using it works? (or is the array copied?)

chqrlie:
> I saw your TODO's in the unit tests. This is a hard question, keep constants in .c2i/.h files and inline them for .c files? For the arrays, since we now have the array and not the pointer in the interface, the user can do `elemsof()`, so doesn't need any other number for that.

The array length is not necessarily documented in the interface file.  For example the lengths of the trace arrays are not known until the code generation is complete so the interface file `c2_trace_tables` purposely use `[]` to declare the arrays without a known length.  Other libraries might have the same needs, eg: the arrays of `FILE` structures in `<stdio.h>` on OpenBSD:
```
typedef struct __sFILE {
    ...
} FILE;

extern FILE __sF[];

#define stdin   (&__sF[0])
#define stdout  (&__sF[1])
#define stderr  (&__sF[2])
```
> Zero sized arrays should not be in the interface, so we should indeed add the size explicitly during generation.

There is a difference between declaring an array with a zero length (`[0]`) and an undefined length (`[]`).  We should distinguish these in the analyser and reject `elemsof` on arrays of undefined length.

> Did you test that generating a library with an array in the interface and then using it works? (or is the array copied?)

`c2_trace.c2` uses arrays declared in `libs/c2/c2_trace_tables.c2i` and this works, but I need the length variables to be defined separately because the lengths are not specified in the interface file (and cannot be).

I don't have a test with an actual library produced from c2 files.  Do you have a test case handy for this?

chqrlie:
I added an error for `elemsof` on arrays of unknown length.
Will look into library tests in the `c2examples` repo.

chqrlie:
I tested arrays with libraries in c2examples.  I just pushed a new version with simplified external array declarations.  Please merge this first as c2trace depends on it.

bvdberg:
Could you add unit-tests for all new features (error when elemsof on array of undefined size, etc).

I dont really like the Component.setLanguage(). It's either C2 or not (C then). We will not do stuff for other languages anyways. Maybe we should change this in the manifest, 'native' (=C2) or 'foreign' (!= C2)

for test/interface/incremental_array.c2t, i think the c2i file should also have the array size in the file. The incremental array is only for building, after that it is a known size.

Otherwise I think it's good to go. I tried the test_lib program, it works indeed!

chqrlie:
> Could you add unit-tests for all new features (error when elemsof on array of undefined size, etc).

Done

> I dont really like the Component.setLanguage(). It's either C2 or not (C then). We will not do stuff for other languages anyways. Maybe we should change this in the manifest, 'native' (=C2) or 'foreign' (!= C2)

Done. I use an `is_foreign` boolean set iff `info.language != c2`

> for test/interface/incremental_array.c2t, i think the c2i file should also have the array size in the file. The incremental array is only for building, after that it is a known size.

I agree, but more work needed, left as a TODO

> Otherwise I think it's good to go. I tried the test_lib program, it works indeed!

:)

PR #324, opened 2025-07-10
c-generator: fix string switch code generation

### LINKED ISSUE #323: string-switch does not always generate correct C code
If the first statement in the string switch case is a declaration, the C generator will add braces:

```c2
  case "ADD":
     u16 i = 10;
```

will become:
```2
  case 2: // "ADD" {
    uint16_t i = 10;
  }
```

chqrlie:
* insert brace before comment in case case block has declarations

Fixes #323


bvdberg:
added unit test case + merged

PR #326, opened 2025-07-16
parser: make checkName a Parser type function and simplify call sequence

chqrlie:


bvdberg:
Good idea, it removes some code duplication. I dont see why the arguments are changed to always true/false. instead of p.in_interface. I would like to enforce naming issues in interfaces as well, since it just improves code and we can because of c_name etc

PR #329, opened 2025-07-25
Analyser: refactor error reporting to simplify code

chqrlie:
* add `Checker.conversionError()` to factor many error reports.
* change `ast.getNativeType()` to `ast.getNativeKind()`, return `BuiltinKind`
* add `BuiltinKind.str()` to convert `BuiltinKind` to the corresponding string
* add `Checker.native_kind` to simplify pointer conversion validation


bvdberg:
merged

PR #341, opened 2025-08-16
C-generator: use native C types in C backend for c_xxx types

chqrlie:
* prevent output of type aliases that have a cname attribute that is a native C type
* update libs/c2/c2.c2i so native C types get output as themselves

### DIFF
--- a/compiler/compiler.c2
+++ b/compiler/compiler.c2
@@ -300,6 +300,8 @@ fn void Compiler.build(Compiler* c,
     if (opts.msan) c.addFeature("__MSAN__", "1");
     if (opts.ubsan) c.addFeature("__UBSAN__", "1");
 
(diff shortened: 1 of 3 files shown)


bvdberg:
I'm not sure I understand the issue you're solving here. Did other other types sometimes go wrong in certain cases?

chqrlie:
> Would you rather see 'unsigned char' than uint8_t in the C code? I never use 'unsigned ..' etc when programming C even. Always the uint32_t etc.

This only affects prototypes and code that uses the C type aliases `c_int`, `c_long`... that are meant to be the native C types themselves.  I don't like to see prototypes generated in the C files for the Standard C library functions that do not use the correct C types.

bvdberg:
yes, I agree the standard C functions should match exactly. Ok i'll merge it..

PR #345, opened 2025-08-21
parser: reject reserved words as local names

chqrlie:
* reject C keywords as function parameter and local variable names, as well
  as struct/union member names.
* add tests

### DIFF
--- a/parser/c2_parser.c2
+++ b/parser/c2_parser.c2
@@ -553,6 +553,10 @@ fn VarDecl* Parser.parseParamDecl(Parser* p, bool is_public) {
         if (!p.checkName(name, p.is_interface)) {
             p.error("a parameter name must start with a lower case character");
         }
+        if (p.tok.reserved) {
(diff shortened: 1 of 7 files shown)


chqrlie:
> Yes, this was definitely needed. I like the way you do it by inserting them like keywords. In parser/keywords.c2 there is a long list of commented out keywords. Did you mean for that to stay in?

Yes the list is needed to add the words in the `Info` structure, but they are not keywords, they are reserved words.  Making them keywords prevents the use everywhere, including for example as function names or module names. You use the verb `register` as a type method name and it feels right to allow that.

With the current C backend, these reserved words cannot be used as struct/union member names, function parameter names nor local variable names, so I reject only these cases.  Note however that we could just modify these names in the generator by tacking `__` at the end to fix this issue.

chqrlie:
> I meant the list in comments ... It seems it was a leftover from development

oops! yes left over from computing the max_index. Fixed.

bvdberg:
great! merged

PR #346, opened 2025-08-23
c2i-generator: split C2i interface generator from C generator

chqrlie:
* fix compound statements output
* fix semicolon output
* simplify code
* handle switch statements
* add tests


generator/c2i/c2i_generator.c2
```diff
@@ -16,53 +16,73 @@
...
+    gen.astPool = astPool;
+    gen.out = string_buffer.create(64*1024, false, 3);
+    gen.output_dir = output_dir;
+
+    component.Component* mainComp = comps.getLast();
+
+    // generate external components
+    for (u32 i=0; i<comps.size(); i++) {
+        component.Component* c = comps.get(i);
+        c.visitModules(Generator.on_module, &gen);
```
bvdberg, line 48: I'm not sure how external modules can be exported, or whether we should allow that. It would be a bit weird, but maybe there is a use-case. If you are creating library1 that depends on some library2 (even in the public symbols), than the .c2i file should contain an import of library2. If depending in a non-public way, then the c2i isn't influenced.

generator/c2i/c2i_generator.c2
```diff
@@ -16,53 +16,73 @@
...
     }
+    // generate main module
+    mainComp.visitModules(Generator.on_module, &gen);
+    gen.out.free();
+}
+
+fn void Generator.on_module(void* arg, Module* m) {
+    if (!m.isUsed() || !m.isExported()) return;
```
bvdberg, line 56: If you are exporting a module, it probably is a top-level module that is not 'used'

PR #347, opened 2025-08-26
assert: output more detailed messages

chqrlie:
* output values of non-literal assert expression comparison operands
* output values of non-literal static_assert operands


bvdberg:
old:
static_assert(3, 4);    // @error{static_assert failed, expected 3, got 4}
new:
static_assert(3, 4);    // @error{static_assert failed: 3 == 4}

I think the old message is clearer right? with 3==4 you dont know which was correct?

This patch also grows the c2c binary by 12 KB...

chqrlie:
> old: `static_assert(3, 4); // @error{static_assert failed, expected 3, got 4}` <br>
> new: `static_assert(3, 4); // @error{static_assert failed: 3 == 4}`
> 
> I think the old message is clearer right? with `3==4` you dont know which was correct?

I made the error message consistent with `assert(3 == 4)`.  If both arguments are literals, there is no *correct* argument value, and the assertion that failed is effectively `3 == 4`.  If you write `static_assert(sizeof(S), 4)`, then the message is `static_assert failed: `sizeof(S) == 4, got 3`, maybe it should be `static_assert failed: `sizeof(S) == 4, sizeof(S): 3`.

With this PR, `assert(sizeof(S) == 4);` produces a runtime error `Assertion failed: `sizeof(S) == 4, sizeof(S): 3`

My goal is to make `static_assert` more consistent with the C and C++ equivalent where the assertion is an expression, not 2 separate arguments and consistent with the `assert` statement, producing more explicit error messages.

> This patch also grows the c2c binary by 12 KB...

The code generated for the 345 `assert` statements is slightly larger with the explicit sub-expression strings and computed values.  I can try and reduce the number of strings generated.

chqrlie:
I modified the messages for better consistency:
```
static_assert(3, 4);                    // @error{static_assert failed: values differ: 3 <> 4}
static_assert(0xff, 0b11111110);        // @error{static_assert failed: values differ: 0xFF (255) <> 0b11111110 (254)}
static_assert(sizeof(S), 4);            // @error{static_assert failed: sizeof(S) is 8, expected 4}
static_assert(4, sizeof(S));            // @error{static_assert failed: sizeof(S) is 8, expected 4}
static_assert(sizeof(T), sizeof(S));    // @error{static_assert failed: values differ: sizeof(T) (4) <> sizeof(S) (8)}
```

PR #349, opened 2025-08-26
language: add support for function default parameter values

chqrlie:


bvdberg:
I can think of a lot of different tests, right now only a bit of the happy flow seems unit-tested:
- adding incompatible default value (const char* arg = 123)
- calling a function with default args with 0/1/2 arguments (fn void test1(i32 a =0, i32 b = 1);
- in struct functions
- in function with varargs (=error)
- ..

bvdberg:
fn void test1 (i32 a = 1, i32 b) {}, there should be an error here right?

I think you skipped that because of the named parameters, but that's only a separate patch still under discussion

PR #357, opened 2025-09-21
Parser: simplify type parsing

chqrlie:
* simplify `Parser.parseTypeSpecifier`: always accept qualifier and arrays
* simplify `Parser.parseSingleTypeSpecifier()`: always accept qualifier
* simplify `Parser.parseOptionalArray()`: always accept arrays, remove base and recursion
* simplify `Parser.parseAsType()`, `Parser.parseAsCastType()`, `Parser.isTemplateFunctionCall()`: remove `has_brackets` handling
* add `isArray()` and `isPointer()` type functions for `TypeRef` and `TypeRefHolder` for clarity and flexibility
* this fixes some problems:
  * `sizeof(const char*)` was rejected.
  * `a < i32.max` was interpreted as a template function call
  * `a < b * c > (d + 1)` was interpreted as a template function call (wrong error)
  * `sizeof(Handle[3])` now works for array types

### DIFF
--- a/analyser/module_analyser_function.c2
+++ b/analyser/module_analyser_function.c2
@@ -27,7 +27,7 @@ fn void Analyser.analyseFunction(Analyser* ma, FunctionDecl* fd) {
 
         // check rtype for array-type
         TypeRef* rtype = fd.getReturnTypeRef();
-        if (rtype.getNumArrays()) {
+        if (rtype.isArray()) {
             ma.error(rtype.getLoc(), "functions are not allowed to return array types");
         }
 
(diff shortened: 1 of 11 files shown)


bvdberg:
It looks ok. When doing performance measurements, it makes both the parsing and analysing slower. The analysing goes from ~8.5 ms -> ~9.0 ms. Parsing is only slightly slower (~0.2 ms). Looking at the patch it's not immediately clear what causes this...

chqrlie:
> It looks ok. When doing performance measurements, it makes both the parsing and analysing slower. The analysing goes from ~8.5 ms -> ~9.0 ms. Parsing is only slightly slower (~0.2 ms). Looking at the patch it's not immediately clear what causes this...

I fixed a bug in `Parser.isTemplateFunctionCall` and made some simplifications in `parseAsCastType`.  Try this latest push.

PR #358, opened 2025-09-22
Analyser: improve function flow analysis

chqrlie:
* add `analyseFlow(Stmt* s)` to analyse all exit paths
* rename ambiguous `hasReturn()` as `hasFlowNext()` and fix semantics
* fix erroneous cases of 'missing return statement'
* track more unreachable code cases: `if`, `while`, `for`, `switch`...
* improve warning control: "no-" is an optional prefix to disable the warning
* add $warning for unreachable-code:
  - disabled by default until bootstrap is regenerated
  - enabled by tester for tests
* unknown warning is now a warning, not an error
* fix tests for unreachable code


bvdberg:
merged

PR #360, opened 2025-09-23
Parser: improve token lookahead system

chqrlie:
* move token lookahead queue from `Tokenizer` to `Parser`
* add `Parser.peekToken()` and `Parser.peekToken2()` (reads token value)
* better check for unbalanced parentheses and brackets
* produce proper error for bogus expressions in lookahead phases

### DIFF
--- a/parser/c2_parser.c2
+++ b/parser/c2_parser.c2
@@ -42,12 +42,17 @@ import ctype local;
 import stdio local;
 
 const u32 NumStmtLists = 16;    // max used for c2c is 8
+const u32 MaxLookahead = 64;    // must be power of 2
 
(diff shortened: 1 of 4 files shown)


chqrlie:
> Exactly the change I was expecting. Parsing performance has not changed,

I am happy with the cleaner interface. Glad the performance is not impacted (as expected).

> but thing patch is also struck with the mysterious slowdown (0.5 ms) during analysis even through no code in there has changed. I'll try some things on this side.

You can try changing `MaxLookahead` to a lower value like `32` and see if the change in memory layout impacts the analyser performance... Not that it should, but just testing is interesting.

PR #362, opened 2025-10-04
Analyser: fix array initializer issues

chqrlie:
* accept string literals for 2D `char` array initializers with designators
* reject non integer array designators
* reject more out of range array designators
* rename `ctv_analyser.checkRange` as `ctv_analyser.checkTypeRange` for clarity
* simplify `analyseInitListArray()` and `checkArrayDesignators()`
* add tests


bvdberg:
makes the analyser a bit slower, but also does a bit more..

PR #373, opened 2025-10-08
c2cat: simplify parsing and output, check consistency

chqrlie:
* output tokens with the original spelling                                                                                                                                           
* use Style enum and select on token kind                                                                                                                                            
* add `--color` and `--nocolor` to force/disable colors in output                                                                                                                        
* add `-c` and `--check` to suppress output and just check syntax                                                                                                                    
* show syntax errors to stderr

### DIFF
--- a/tools/c2cat.c2
+++ b/tools/c2cat.c2
@@ -20,8 +20,8 @@ import attr local;
 import color local;
 import file_utils;
 import keywords;
+import src_loc local;
 import string_buffer;
-import string_list;
 import string_map;
 import string_pool;
 import number_radix;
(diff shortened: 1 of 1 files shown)


chqrlie:
c2cat can also be used to just parse the tokens and check for tokenization errors, or even just benchmark the tokenizer...
I will post an update to make this more obvious.

bvdberg:
If I run 'c2cat <file.c2> --nocolor'  I still get colored output...

chqrlie:
I fixed that: options can be specified before or after filenames.
also recognize `--` to stop further option parsing.
exit status is non zero if any parsing error is detected.

PR #383, opened 2025-10-18
Compiler: force zero initialization of local variables

chqrlie:
* make all local variables zero-initialized except if tagged as `@(noinit)`
* update tests

### DIFF
--- a/ast/ast_evaluator.c2
+++ b/ast/ast_evaluator.c2
@@ -408,7 +408,7 @@ fn Value Evaluator.eval_call(Evaluator* caller, const CallExpr* c) {
 
     // Create a new stack frame and link it to the caller's
     // TODO: handle Stack frames as separate allocated objects or fragments of a stack area
-    Evaluator eval;
+    Evaluator eval /*@(noinit)*/;
 
     if (num_args > elemsof(eval.args)) {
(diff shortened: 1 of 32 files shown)


chqrlie:
> It should easily opt out zero init, already at -O1. This works fine here at least: https://godbolt.org/z/M4sTEr5zv

Redundant initialization seems easy to handle.

How about this one:

```
module test;
import stdio local;

fn void main() {
    char buf[4096];
    snprintf(buf, sizeof(buf), "Hello world");
    printf("%s\n", buf);
    return 0;
}
```

chqrlie:
> That's something that would require `@noinit`. Note that it would not be correct to elide the implicit zeroing in this case, since the bytes not affected by `snprintf` will not be modified, and that might be what the programmer intended.

Indeed `snprintf` does not affect the bytes beyond offset 11, but since `printf` never reads them, leaving them uninitialized is a palatable optimisation.  This is very similar to the case of unused local variables: I suppose C3 does not initialize them either.

bvdberg:
I tested this, performance loss seems none (even without the noinit flags commented out because of the bootstrap). I just miss some tests that test this noinit flag. Otherwise it's fine to merge.

PR #385, opened 2025-10-21
Compiler: add warning control over `max_identifier_length`

chqrlie:
* pass warning flags to parser and tokenizer
* let specific projects allow longer identifiers with `$warnings no-max-identifier-length`

### DIFF
--- a/common/build_target.c2
+++ b/common/build_target.c2
@@ -203,6 +203,7 @@ public fn void Target.disableWarnings(Target* t) {
     t.warnings.no_unreachable_code = true;
     t.warnings.no_unknown_attribute = true;
     t.warnings.no_deprecated = true;
+    t.warnings.no_max_identifier_length = true;
 }
 
 public fn void Target.enableWarnings(Target* t) {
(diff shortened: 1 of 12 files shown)


chqrlie:
I am closing this PR because we should focus first on how we construct C identifiers.  If we remove the `MaxIdentifierLen` limit, we should be more cautious and reject some combinations that exceed the limit for C identifier length.

The current C Standard (C23) mandates these minimums for conformant compilers:
- 63 significant initial characters in an internal identifier or a macro name.
- 31 significant initial characters in an external identifier.

These limits are very conservative and implementations are encouraged to go well beyond:

**6.11 Future language directions**<br>
**6.11.3 External names**<br>
Restriction of the significance of an external name to fewer than 255 characters (considering each
universal character name or extended source character as a single character) is an obsolescent feature
that is a concession to existing implementations.

So we should limit the length of C identifiers produced to 255 characters and allow up to this limit.  The buffers currently used for this have a length or 64 or 128 bytes.  I shall investigate and enforce better consistency.

PR #389, opened 2025-10-31
Ast: improve consistency in instantiators

chqrlie:
* add missing stats
* use direct return statements with casts


bvdberg:
nice!!

PR #391, opened 2025-11-02
Compiler: use enum based color handling

chqrlie:
* add `Color` enum and constant aliases
* define standard colors as enum values

### DIFF
--- a/ast/utils.c2
+++ b/ast/utils.c2
@@ -16,7 +16,7 @@
 module ast;
 
 import attr;
-import color;
+import color local;
 import ast_context local;
 import string_buffer;
 import string_pool;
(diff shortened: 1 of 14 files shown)


ast_utils/string_buffer.c2
```diff
@@ -126,8 +128,8 @@ public fn void Buf.clear(Buf* buf) {
     buf.truncated = false;
 }
 
-public fn void Buf.color(Buf* buf, const char* color) {
-    if (buf.colors) buf.add(color);
+public fn void Buf.color(Buf* buf, Color col) {
+    if (buf.colors) buf.add(col.str());
```
bvdberg: Inside col.str(), there is another check for if colors. Maybe have a separate str() function for use outside StringBuffer?

ast/utils.c2
```diff
@@ -307,17 +307,17 @@ public fn BuiltinKind getNativeKind() {
...
+Color col_Decl = Color.Bgreen;
+Color col_Expr = Color.Bmagenta;
+Color col_Attr = Color.Blue;
+Color col_Template = Color.Green;
+//Color col_Cast = Color.Red;
+Color col_Type = Color.Green;
+Color col_Value = Color.Bcyan;
+Color col_Error = Color.Red;
+Color col_Calc = Color.Yellow;    // all calculated value
+Color col_Normal = Color.Normal;
```
bvdberg: You can actually leave out the Color. in the right-hand side.. :)
Color col_Stmt = Bmagenta;

ast_utils/color.c2
```diff
@@ -16,25 +16,73 @@
...
+    [Color.Bcyan]    = "\033[1;36m",
+    [Color.White]    = "\033[1;37m",
+    [Color.Normal]   = "\033[0m",
+}
+
+bool hasColor;
 
 public fn bool useColor() {
-    return unistd.isatty(1);
+    return hasColor = unistd.isatty(1);
```
bvdberg: ehhh this line does not parse well for me.. I think making it 2 lines is more readable.
hasColor = ..;
return hasColor;

ast_utils/string_buffer.c2
chqrlie: Fixed: `col.str()` always returns the string associated with `col`.

ast_utils/color.c2
chqrlie: I removed the `hasColor` global variable.

PR #393, opened 2025-11-05
Compiler: add local function statements and expressions

chqrlie:
* local functions are defined using the standard syntax in the body of a function.
* they cannot be templates nor type functions
* they can be named or anonymous
* accept a capture list between square backets fater the parameter list (currently ignored)


chqrlie:
This branch is not ready to merge yet.  I am closing the PR for now.  The main issue is local functions cannot refer to identifiers from the scope where the definition is present.  This is a tricky issue if we want reliable and reasonably safe operation.

PR #400, opened 2025-11-29
Literals: improve literal type consistency

chqrlie:
* insert implicit casts on constant expressions
* pass `BuiltinKind` to numeric literal ast constructors
* determine `BuiltinKind` according to C semantics: `2147483648` is an `i64` while `0x80000000` is a `u32`
* use literal's `BuiltinKind` to determine C output syntax
* use uppercase suffixes for readability
* fix c2 library definitions so there are no implicit conversions
* fix Makefile so trace targets do not pollute output/c2c
* add tests on literal `sizeof` expressions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,13 +73,13 @@ output/c2c_trace/c2c_trace: $(C2C_DEPS)
 	$(C2C) c2c --trace-calls -o c2c_trace --fast
 
 trace_calls: $(C2C) output/c2c_trace/c2c_trace
-	C2_TRACE="min=10;min2=1;mode=3;name=*;fd=2" output/c2c_trace/c2c_trace c2c --test 2> output/c2c/calls
+	C2_TRACE="min=10;min2=1;mode=3;name=*;fd=2" output/c2c_trace/c2c_trace c2c -o c2c_calls --test 2> output/c2c/calls
 
 trace: trace_calls
 	cat output/c2c/calls
 
 alloc_trace: $(C2C) output/c2c_trace/c2c_trace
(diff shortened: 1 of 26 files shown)


bvdberg:
The following piece of code seems to give an error:
fn void test1(f32 f) {
    f32 a = f * 0.6;    // error: implicit conversion loses floating-point precision: 'f64' to 'f32'
}

That does not seem right...

chqrlie:
Indeed this is a case of floating point precision loss in C because `0.6` is a `double`, not a `float`, so the expression `f * 0.6` is evaluated as `(double)f * 0.6` and the result is implicitly converted back to `(float)` upon assignment to `f`.

Do you propose that `0.6` be typeless until used in an expression or assigned, thus implicitly converted to `float` if used in a `float` binary expression or used in a `float` context (rhs of an assignment, passed as an argument, returned...)?

This feels intuitive and seems to improve over often misunderstood C semantics, and equivalent to an implicit `F` suffix in the equivalent C expression.

This does not fix all cases: CTVs are still evaluated as `double`:
* `f = f * -0.6;`
* `f = f * (1.0 / 6);`

A general fix for this is to insert an implicit `(float)` conversion on CTVs, not just floating point literals.  This implicit conversion would be made explicit in the C code generator to avoid warnings and ensure the above semantics.

chqrlie:
I removed this change in the analyser to simplify merging this patch and only handle floats.

I have another pull request pending with a more extensive rewrite of **module_analyser_binop.c2** for argument checking and argument conversions.

PR #409, opened 2025-12-13
Analyser: construct new format strings in analyser

chqrlie:
* add `AlternateExpr` nodes
* compute modified format string in analyser and store via `Alternate` node
* move analyser_utils/printf_utils.c2 to analyser/printf_utils.c2
* remove unused code
* fix floating point formats: `%lf` is not required for `printf`

### DIFF
--- a/analyser/module_analyser_call.c2
+++ b/analyser/module_analyser_call.c2
@@ -19,8 +19,10 @@ import ast local;
 import printf_utils;
 import scope;
 import src_loc local;
+import string_buffer;
 
 import stdio;
+import string;
 
 const char[] DiagTooManyArgs = "too many arguments to %sfunction call, expected %d, have %d";
 const char[] DiagTooFewArgs = "too few arguments to %sfunction call, expected %d, have %d";
@@ -163,7 +165,7 @@ fn QualType Analyser.analyseCallExpr(Analyser* ma, Expr** e_ptr) {
 
     //stdio.printf("CALL (%d | %d)\n", func_num_args, call_num_args);
     bool has_printf_format = false;
-    u32 printf_call_idx = 0;
+    u32 format_arg_idx = 0;
     while (1) {
         //stdio.printf("ARG [%d | %d]\n", func_arg_index, call_arg_index);
         if (func_arg_index >= func_num_args) break;
@@ -212,7 +214,7 @@ fn QualType Analyser.analyseCallExpr(Analyser* ma, Expr** e_ptr) {
 
         if (vd.hasPrintfFormat()) {
             has_printf_format = true;
-            printf_call_idx = call_arg_index;
+            format_arg_idx = call_arg_index;
         }
 
         func_arg_index++;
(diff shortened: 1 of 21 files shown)


ast/alternate_expr.c2
```diff
@@ -0,0 +1,74 @@
...
+
+import ast_context;
+import string_buffer;
+import src_loc local;
+
+public type AlternateExpr struct @(opaque) {
+    Expr base;
+    Expr* inner;
+    Expr* generated;
+}
```
bvdberg: Maybe rename inner/getInner to original/getOriginal? When reading the code, I wondered what inner was

analyser/module_analyser_call.c2
```diff
@@ -357,39 +402,48 @@ fn bool on_format_specifier(void* context, printf_utils.Specifier specifier, u32
...
+    SrcLoc format_loc = format.getLoc();
     const char* format_text = printf_utils.get_format(format, &format_loc);
-    if (!format_text) goto not_constant;
+    if (!format_text) {
+        ma.error(format_loc, "format argument is not a constant string");
+        return false;
+    }
+
+    FormatAnalyser fa = { ma, format_text, args, num_args, 0, format_loc, 0 }
+    fa.out = string_buffer.create(256, false, 0);
```
bvdberg: Put these 2 lines in FormatAnalyser.init()? removes all the constants here. Also add FormatAnalyser.free() to hide more internals. What do you think about adding a single FormatAnalyser in the ModuleAnalyser as member?

bvdberg:
Otherwise very nice patch! It lays the foundation for more changes  (CTV etc)

generator/ast_visitor_expr.c2
```diff
@@ -119,6 +114,11 @@ fn void Visitor.handleExpr(Visitor* v, Expr* e) {
         NamedArgument* n = (NamedArgument*)e;
         v.handleExpr(n.getInner());
         break;
+    case Alternate:
+        AlternateExpr* n = (AlternateExpr*)e;
+        v.handleExpr(n.getInner());
+        v.handleExpr(n.getGenerated());
+        break;
```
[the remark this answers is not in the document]
chqrlie: Yes, only the inner node should be visited.  Where would you move this module?

analyser/module_analyser_call.c2
```diff
@@ -265,18 +267,57 @@ fn QualType Analyser.analyseCallExpr(Analyser* ma, Expr** e_ptr) {
...
+    [BuiltinKind.UInt16] = false,
+    [BuiltinKind.UInt32] = false,
+    [BuiltinKind.UInt64] = true,
+    [BuiltinKind.Float32] = false,
+    [BuiltinKind.Float64] = true,
+    [BuiltinKind.ISize] = true,
+    [BuiltinKind.USize] = true,
+    [BuiltinKind.Bool] = false,
+    [BuiltinKind.Void] = false,
+}
```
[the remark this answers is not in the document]
chqrlie: Done: renamed as `Format_size_prefix`

analyser/module_analyser_call.c2
chqrlie: I simplified the code.

I shall improve this part in a later patch, along with support for `scanf` format string analysis and conversion.

bvdberg:
For a simple printf, the AST now becomes:
    Alternate const char* nvalue
     ImplicitCast const char* CTC lvalue ArrayToPointerDecay
      StringLiteral const char[7] CTC lvalue "%d %d\n"
     StringLiteral const char[8] CTC lvalue "%d %lu\n"

So the original stringliteral is wrapped in a ArrayToPointerDecay, but the new one isn't... I would assume the new one would be wrapped, but the old one wouldn't

PR #419, opened 2025-12-30
Language: attr deprecated attribute

chqrlie:
* add `@(deprecated="msg")` attribute for functions that should not be used
* mark library functions `gets()`, `strncpy()` and `strncat()` as deprecated
* replace bogus call to `strncpy()` with `pstrcpy()`

### DIFF
--- a/analyser/module_analyser_call.c2
+++ b/analyser/module_analyser_call.c2
@@ -16,6 +16,7 @@
 module module_analyser;
 
 import ast local;
+import attr local;
 import printf_utils local;
 import scope;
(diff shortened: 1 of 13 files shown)


chqrlie:
> this will be handy to have. Could you also change the FunctionDecl.print() to show the attribute if printed?

Done

> Also I think we need a new warning category for this, that you can disable (enabled by default). So change from error to warning?

I added a new `deprecated` warning category, disabled with`-Wno-deprecated`, and added more tests for this.

PR #441, opened 2026-02-14
bootstrap: reduce bootstrap size by diabling IR, stats, asserts

chqrlie:
* add `--bootstrap` compiler option
* add `$optional` recipe target flag
* add bootstrap target with BOOTSTRAP feature
* make EXPERIMENTAL features standard
* update bootstrap code


bvdberg:
We have a cmdline arg '--help-recipe' that prints out all the recipe options (live documenation ;) ). Could you add the $optional to that as well?

PR #444, opened 2026-02-14
string_buffer: simplify and optimize

chqrlie:
* only store null terminator when getting data pointer
* make `Buf.data()` return `const void*`
* remove `Buf.udata()`, add `const char* Buf.str()`


bvdberg:
merged!

PR #448, opened 2026-02-19
Libs: split C types into c_types module

chqrlie:
* c2 module only has min_xxx and max_xxx definitions
* c_types module has the C type aliases


bvdberg:
merged

PR #449, opened 2026-02-21
Compiler: fix type/expr ambiguity for builtin min/max properties

chqrlie:


bvdberg:
Yeah nice.. I think we can remove the old max_u32 etc constants now (there are 2 usages in ir/ somewhere, but they can be converted to u32.max)

PR #471, opened 2026-03-21
Compiler: improve C native type parser

chqrlie:
* support function pointer type definitions


bvdberg:
yeah a bit cleaner again! merged

PR #473, opened 2026-03-22
Language: add `when` statement to select code and definitions

chqrlie:
* `when` has the same syntax as `if` but with these restrictions:
  * condition must be a compile time constant value
  * if true, only the *then* branch is analysed and generated
  * otherwise, only the *else* branch, if present, is analysed and generated
  * `when` blocks do not start a new scope
  * `when` cannot appear as the sole statement in `if`, `else`, `for` or `while`


chqrlie:
This branch is WIP.  I will reopen a PR when templates are ready to merge and `when` blocks are available at the global scope, inside a `template` block.

PR #483, opened 2026-03-31
Language: add `tlocal` keyword for thread local storage class

chqrlie:
* allow local and global variables with thread local storage class
* most `local` variables in the compiler should be made `tlocal`.


bvdberg:
merged!

PR #484, opened 2026-04-01
Language: add labelled break and continue statements

chqrlie:
* `break` and `continue` can take the label of an enclosing statement to break or continue directly, bypassing the nested code flow.
* simplified flow analysis
* add tests

### DIFF
--- a/ast/break_stmt.c2
+++ b/ast/break_stmt.c2
@@ -19,21 +19,43 @@ import ast_context;
 import string_buffer;
 import src_loc local;
 
-public type BreakStmt struct @(opaque) {
-    Stmt base;
+type BreakStmtBits struct {
+    u32 : NumStmtBits;
+    u32 has_name: 1;
 }
 
-public fn BreakStmt* BreakStmt.create(ast_context.Context* c, SrcLoc loc) {
-    BreakStmt* s = c.alloc(sizeof(BreakStmt));
+public type BreakStmt struct @(opaque) {
+    Stmt base;
+    u32[0] name_idx;
+}
+
+public fn BreakStmt* BreakStmt.create(ast_context.Context* c, SrcLoc loc, u32 name_idx) {
+    u32 size = sizeof(BreakStmt);
+    if (name_idx) size += sizeof(u32);
+    BreakStmt* s = c.alloc(size);
     s.base.init(Break, loc);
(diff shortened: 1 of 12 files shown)


bvdberg:
I'm trying to find a syntax that does not look like a label (since it's not really a label), what about:
```
fn i32 test1(i32 a, i32 b) {
    while [outer] (a > 0) {
        while [inner] (b > 0) {
            switch [switch1] (b) {
            case 1:
            case 2:
                break outer;
            case 3:
                switch (a + b) {
                case 10:
                    continue outer;
                case 20:
                    continue inner;
                case 30:
                    break switch1;
                default:
                    break inner;
                }
                break;
            }
        }
    }
}
```

So after a for/while/switch keyword we can have a loop name between square brackets, like [name].
A continue can only be used on loops.

chqrlie:
While your idea has some merit, the syntax adopted by the C committee is what it is and follows the precedent of other languages (javascript, java, C++, C#, Perl, etc.) with the same semantics.  We should use the same approach and semantics.  The label in front of the statement can be used for `goto`, `break` and `continue` as expected. 

Regarding style, the label is usually set on the same line as the `for`, `while`, `switch` it labels:
```
fn i32 test1(i32 a, i32 b) {
    outer: while (a > 0) {
        inner: while (b > 0) {
            switch1: switch (b) {
            case 1:
            case 2:
                break outer;
            case 3:
                switch (a + b) {
                case 10:
                    continue outer;
                case 20:
                    continue inner;
                case 30:
                    break switch1;
                default:
                    break inner;
                }
                break;
            }
        }
    }
}
```
I am not a big fan of this, but it seems to be better received than the `goto` statements we tend to use for the same purpose.

ast/break_stmt.c2
```diff
@@ -19,21 +19,43 @@ import ast_context;
...
 import src_loc local;
 
+type BreakStmtBits struct {
+    u32 : NumStmtBits;
+    u32 has_name: 1;
+}
+
 public type BreakStmt struct @(opaque) {
     Stmt base;
+    u32[0] name_idx;
```
bvdberg: We could use 28 bits in the BreakStmtBits to store the name. Value 0 would mean to name. Then no tail allocated stuff would be needed. What do you think?

PR #487, opened 2026-04-02
Language: split decl and cond in `IfStmt`

chqrlie:
* simplify code generation for `if` statement with declaration
* support for optional separate decl and cond as per next C Standard


bvdberg:
merged

PR #488, opened 2026-04-04
Compiler: refine console configuration levels and share with plugins

chqrlie:
* add `-q` and `--quiet` c2c option to disable console.log messages
* add `console.setQuiet` to disable `console.log` messages
* collect console configuration in a `Config` object
* pass console configuration to plugins
* change quiet flag in tester from `-t` to `-q` (for consistency)
* make `Makefile` rebuild targets and examples less verbose

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,11 @@ $(C2C): output/bootstrap/bootstrap $(C2C_DEPS)
 	@rm -rf output/bootstrap/cgen
 	@mv -f output/c2c/* output/bootstrap
 	@echo "---- running c2c (no plugins$(C2FLAGS)) ----"
-	@output/bootstrap/c2c $(C2FLAGS) --noplugins --fast c2c $(PLUGINS)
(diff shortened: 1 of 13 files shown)


bvdberg:
For tooling wrapping the compiler this makes it harder to automate tasks. I prefer that output is always the same, so that tooling can easily grep stuff.

chqrlie:
Fair enough... How about a command line option `--terse` that prevents the output of `building xxx`?

bvdberg:
I've never seen 'terse' used, maybe use 'quiet'. In console, we already check console.debug() for some flag. Maybe this is in line with that, but it also suppresses console.log? Then we only need to check in one place.

chqrlie:
> I've never seen 'terse' used, maybe use 'quiet'. In console, we already check console.debug() for some flag. Maybe this is in line with that, but it also suppresses console.log? Then we only need to check in one place.

Actually the tester uses `-t`, described as `terse mode`...

I changed the flag to `-q` / `--quiet` and grouped the console configuration flags in a `console.Config` object shared with the plugins.  The color customization could be stored there too later.

bvdberg:
merged

PR #489, opened 2026-04-09
Compiler: use '\0' for null bytes for consistency

chqrlie:


bvdberg:
I always think of '\0' as a sort of ani-pattern. The value is just 0. For \r \n it's easier to read than the (hex/decimal)  value, but for value 0, it makes it harder to read IMHO.

chqrlie:
You are correct, the value is `0`, but I like to make it more obvious in source code which version of `0` is used:
- `'\0'` for characters
- `0.0` for floating point comparisons
- `NULL` for pointers

It is indeed a question of style and I respect your disagreement.

PR #491, opened 2026-04-09
Analyser: generate alternate nodes for case strings

chqrlie:
* generate alternate nodes for case conditions at analysis time
* simplify generators for switch cases
* fix switch on string with combined cases
* add test case
* fix `fread` prototype

### DIFF
--- a/analyser/module_analyser_switch.c2
+++ b/analyser/module_analyser_switch.c2
@@ -159,35 +159,44 @@ fn FlowBits Analyser.analyseSwitchStmt(Analyser* ma, Stmt* s) {
     // TODO: handle complete coverage
     if (!defaultCase) flow |= FlowNext;
 
-    if (!(flow & FlowError) && is_string) {
+    if (!(flow & FlowError) && is_string && !ma.check_only) {
         // Create alternate expression for c2_strswitch(s)
         // create call expression `c2_helper.c2_strswitch("expression", "case string")
         // TODO: handle declaration expressions
         string_buffer.Buf* casestr = string_buffer.create(128, false, 0);
+        u32 str_index = 2;
 
         for (u32 i = 0; i < numCases; i++) {
             SwitchCase* c = cases[i];
             if (c.isDefault())
                 continue;
-            Expr* e = c.getCond(0);
-            if (e.isImplicitCast()) {
-                const ImplicitCastExpr* ic = (ImplicitCastExpr*)e;
-                e = ic.getInner();
-            }
-            if (e.isNil())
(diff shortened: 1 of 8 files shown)


bvdberg:
When calling c2c --check, no alternate nodes need to be generated.

When trying this on a simple case:
```
fn i32 test1(const char* input) {
    switch (input) {
    case "aa":
        return 1;
    case "bb":
        return 2;
    case "cc":
        break;
    }
    return 4;
}
```

The AST looks like:
```
FunctionDecl i32 (const char*) unused Normal def:global test1
 VarDecl const char* parameter input
 CompoundStmt
  SwitchStmt string
   Alternate i32 rvalue
    ImplicitCast const char* rvalue LValueToRValue
     Identifier const char* lvalue Var input
    Call i32 rvalue
     Member i32 (const char*, const char*) CTC rvalue Module refs=2/2 c2_helper.c2_strswitch
     ImplicitCast const char* rvalue LValueToRValue
      Identifier const char* lvalue Var input
     ImplicitCast const char* CTC lvalue ArrayToPointerDecay
      StringLiteral const char[10] CTC lvalue "\002aa\002bb\002cc"
   SwitchCase
    Alternate const char[3] nvalue
     StringLiteral const char[3] CTC lvalue "aa"
     IntegerLiteral i32 CTC CTV rvalue 2
     ReturnStmt
      IntegerLiteral i32 CTC CTV rvalue 1
   SwitchCase
    Alternate const char[3] nvalue
     StringLiteral const char[3] CTC lvalue "bb"
     IntegerLiteral i32 CTC CTV rvalue 3
     ReturnStmt
      IntegerLiteral i32 CTC CTV rvalue 2
   SwitchCase
    Alternate const char[3] nvalue
     StringLiteral const char[3] CTC lvalue "cc"
     IntegerLiteral i32 CTC CTV rvalue 4
     BreakStmt
  ReturnStmt
   IntegerLiteral i32 CTC CTV rvalue 4
```

Wouldn't it be more convenient to just have 1 alternate Stmt of the whole switch?
Right now there seems to be one for each case + one for the condition.

chqrlie:
Sure `--check` could avoid constructing the alternate nodes (done).

We could use an alternate statement but there is no such `Stmt` kind yet and it would also duplicate the case nodes and the statement arrays, which would be even more wasteful.  String switches are not so common anyway.
As an alternative, we could use a new `ConstantExpr` node that would hold the value and an inner `Expr*`.  This would be useful for constant integer expressions that are evaluated at analysis time and used at generation time.

PR #495, opened 2026-04-13
Compiler: make `setTypeUsedPublic()` a type function

chqrlie:
* move function `setTypeUsedPublic()` from ast/utils.c2 to ast/qual_type.c2 as type function `QualType.setUsedPublic()`
* this function is really a type function and does not depend on `globals`


bvdberg:
merged

PR #501, opened 2026-04-15
Compiler: remove many calls to ast.idx2name

chqrlie:
* add `idx2name` converter in `unused_checker.Checker`
* add `Generator.emitName` in c2i_generator and c_generator
* remove most calls to `Decl.getName` in generators
* remove unused functions


bvdberg:
This pattern now happens quite a lot: `ma.idx2name(d.getNameIdx()))` Maybe add a ma.getName(Decl* d) that does just that?

chqrlie:
I came to the very same conclusion this morning... and `ma.getFullName` too :)

<strike>But lets make it a separate patch</strike>

Never mind, I shall add and use new wrappers

PR #503, opened 2026-04-16
Compiler: control more C compiler warnings via pragmas in the `_external.h` header.

chqrlie:


bvdberg:
What would be the advantage in doing it that way?

chqrlie:
Multiple advantages:

- simpler command line to invoke the C compiler in the (generated) `Makefile`
- more portable C compiler command line: warnings vary from one compiler to another
- group all warning control in the same place

PR #505, opened 2026-04-17
AST: simplify uses of ast_idx

chqrlie:
* make `idx2ast` and `addAst` type methods of `globals`
* ensure `globals.ast_list[0]` is `nil` and simplify `idx2ast()`
* remove redundant `ast` prefix in ast/module.c2 and
* remove `Decl.getASTIdx`


bvdberg:
yeah nice! merged

PR #508, opened 2026-04-18
Compiler: simplify word size testing

chqrlie:
* add `native_kind` in `ast_builder.Builder`
* group allocation stuff in `TailAllocator` as a member of `ast_context.Context`
* add `wordsize` in `ast_context.Context`
* move compiler sub-object creation after target analysis
* use `gen.targetInfo` for word size in C Generator
* add `align` member in `Type.typebits`, use for builtins, pointers and functions
* initialize dummy slot 0 in pointer pool
* move `getWordSize()` and `getNativeKind()` from global to ast_context
* simplify `size_analyser.sizeOfType`: no reference to `global` needed
* conversion_checker gets `native_kind` from builder/context
* simplify `getUnaryOpWidth`, no global reference anymore

### DIFF
--- a/analyser/conversion_checker.c2
+++ b/analyser/conversion_checker.c2
@@ -118,7 +118,7 @@ public type Checker struct {
 public fn void Checker.init(Checker* c, diagnostics.Diags* diags, ast_builder.Builder* builder) {
     c.diags = diags;
     c.builder = builder;
-    c.native_kind = ast.getNativeKind();
+    c.native_kind = builder.getNativeKind();
     // leave other fields uninitialized
 }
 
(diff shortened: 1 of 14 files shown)


bvdberg:
What's the purpose of splitting Context into a TailAllocator and a Context?

chqrlie:
The `TailAllocator` belongs in `ast_utils` and should only deal with allocation

The `ast_context.Context` should hold the core ast data, such as the builtins, the string and pointer pool, and some ancillary data such as the wordsize. It should be moved to the `ast` module as `ast.AstContext` and passed to the different modules.

This PR performs an initial step in this direction but it might be better to make a more radical move, for which I have a larger patch in progress.

bvdberg:
merged

PR #510, opened 2026-04-18
Compiler: simplify ast statistics report

chqrlie:
* add `ast.report` and rename `Stats.dump` as `Stats.report` for consistency
* report statistics explicitly instead of as a side effect of `ast.deinit()`
* report statistics for `global` arrays
* improve ast stats report: factor code, show average size


bvdberg:
merged

PR #511, opened 2026-04-19
Compiler: remove `enum_min` and `enum_max` builtin operators

chqrlie:
* remove `enum_min` and `enum_max` keywords and expressions
* use u64 instead of Value type for computed value of builtin expressions
* add `use_literal` option for C generator (WIP)
* simplify code
* update tests


bvdberg:
nice, including updating the documentation!

PR #513, opened 2026-04-21
C-Generator: fix enum assoc values dependency analysis

chqrlie:
* handle `EnumConstant` in dep_finder.c2 and `c_generator.emitGlobalDecl`
* add test case with recursive enum dependency
* this test failed in previous builds unless files were specified in the opposite order for compilation


bvdberg:
merged

PR #516, opened 2026-04-21
Ast: store length of string literal instead of size

chqrlie:
* add `StringLiteral.getLength()` to get the length in bytes of a string literal ast node, and make this the stored value it is used more often than the size, which includes the trailing null terminator
* add TODOs in IR generator regarding embedded null bytes in string literals


bvdberg:
merged

PR #517, opened 2026-04-21
String-buffer: simplify `string_buffer.Buf.encodeBytes` API

chqrlie:
* `encodeBytes` always outputs the surrounding separtors if passed a `sep` character and escapes both `"` and `'` if `sep == 0`.


bvdberg:
merged

PR #518, opened 2026-04-22
Parser: add `VarDeclList` for function parameters

chqrlie:
* add `VarDeclList` with the exact same code as `DeclList`
* use `data` and `getData()` for genericity
* use `stdlib.realloc()` to simplify code
* remove casts


bvdberg:
merged

PR #519, opened 2026-04-22
string_pool: allocate by blocks to ensure string pointers remain valid

chqrlie:
* instead of a single reallocated block, use an array of blocks                                                                                                                      
* this ensures that all pointers returned by `idx2str` remain valid until                                                                                                            
  the string pool is freed.

### DIFF
--- a/ast_utils/string_pool.c2
+++ b/ast_utils/string_pool.c2
@@ -1,4 +1,4 @@
-/* Copyright 2022-2026 Bas van den Berg
+/* Copyright 2022-2026 Bas van den Berg, Charlie Gordon
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


bvdberg:
Having a block stash seems like a bit of overkill right? This thing allocated blocks of 16KB, so simplifying the code and removing the stash seems better. What is your opinion? It does slow down the compiler a bit (~ 2%)

chqrlie:
I guess the stash is not needed.  I added it so the Pool use a single allocation upon creation and before I realized the block could not be larger than `BLOCK_SIZE` the stash would have covered most cases.
I am going to investigate another solution that should also address the performance hit.  Is the 2% you measure in the parser or elsewhere?

bvdberg:
The slowdown appears divided over parser and analysing. Even when starting with a 32 KB astPool, it is still not slower with the original code

chqrlie:
I pushed a simpler implementation without a stash. Can you check the performance?

bvdberg:
UPDATE: apparently it was an out-of-date plugin. works.

I tested it without the stash and that performs well. I was about to merge it and ran one more test on my game engine project, that failed:
```
...
generating refs
generator/c2refs.c2:217: function c2refs.Files.start: assertion failed: idx != NOT_FOUND, idx: 4294967295, NOT_FOUND: 4294967295
Aborted (core dumped)
```
I did a clean build there. Not sure what caused it. Lets look into it

bvdberg:
merged

chqrlie:
We should work on a plugin API signature system to detect incompatible plugins at load time...
Maybe we should internalize the git version and use that as the signature.

PR #521, opened 2026-04-24
Examples: remove redundant casts

chqrlie:
* use `void*` for `file.data()` and remove `file.char_data()`
* remove many redundant casts


bvdberg:
merged!

PR #523, opened 2026-04-27
Analyser: simplify `Analyser.analyseMemberExpr`

chqrlie:
* remove loop in analyser as `MemberExpr` always has 2 elements
* further simplifications are pending


bvdberg:
merged

PR #532, opened 2026-07-09
Analyser: detect and report invalid assignment in assert

### LINKED ISSUE #529: assert issue not detected
assert(a = b); gives to error, while it should be (assert == b);
Now it generates code to assign b to a

chqrlie:
* cannot have an assignment as an assert expression

fixes #529


bvdberg:
nice!

PR #533, opened 2026-07-09
Compiler: refine support of memory models (ILP32, LP64, LLP64)

### LINKED ISSUE #499: C compilation errors on builtin module
The functions in the `builtin` module produce a C compilation error as they are defined to return `u8`, but the C `__builtin_xxx` equivalents return `int`. The forward declaration produced results in a type mismatch.

Minimal example to reproduce:
```
module example;
import builtin as b;

public fn i32 main(i32 argc, char **argv) {
    return b.popcount(cast<u32>(argc));
}
```

This gives an error with clang and a warning and link error with GCC.

chqrlie:
* renamed `target_info.Info.intWidth` as `ptrWidth`
* add `target_info.Info.longWidth` distinct from `target_info.Info.ptrWidth` to support LLP64 architectures. see https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models
* define `LONG_32BIT` and `LONG_64BIT`, distinct from `ARCH-32BIT` and `ARCH_64BIT`
* fix builtin function prototypes

fixes #499

### DIFF
--- a/common/target_info.c2
+++ b/common/target_info.c2
@@ -87,7 +87,8 @@ public type Info struct {
     Vendor vendor;
     Abi abi;
 
-    u32 intWidth;   // 32/64
+    u32 ptrWidth;   // 32/64
+    u32 longWidth;  // 32/64
 
     char[80] triple;
 }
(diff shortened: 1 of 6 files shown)


bvdberg:
Looking at the Wiki only LP64 seems to be relevant right? Why would we support any other?

PR #534, opened 2026-07-09
Compiler: use enum type as array sizes if possible

chqrlie:
* use enum name instead of `elemsof(type)` for arrays that are indexed by enum values.


chqrlie:
Merged!

PR #538, opened 2026-07-11
Github-CI: add linux 32-bit target (DO NOT MERGE)

chqrlie:
* enable linux 32-bit target                                                                                                            
* fix 32-bit issues                                                                                                                     
* add 32-bit struct sizes in `static_assert` validations                                                                                
* create **bootstrap/bootstrap-linux-i686.patch**


bvdberg:
I think the ```static_assert(sizeof(void*) == 4 ? 132 : 144+64, sizeof(AST));``` is an anti-pattern.
It would be better to extend the static_assert syntax to support this natively.

Options I can think of:
 
static_assert(8, sizeof(void*));  // generic case
static_assert[32](4, sizeof(void*));  // 32 bit platforms only
static_assert[64](8, sizeof(void*)); // 64 bit platforms only

// have different static assert functions (also the generic one)
static_assert32(..);
static_assert64(..);

// if needed (not now) we could extend the first syntax to include arch/system limits. The list of these is already in common/target_info.c2.
static_assert[arch=arm64](..)
static_assert[arch=arm64,system=darwin](..)

I think the arch/system options might be needed in the future, so perhaps the [] syntax is the better choice..

PR #544, opened 2026-07-22
Examples: rewrite example/sudoku into a full solver

chqrlie:
* optimize direct resolution method
* add brute force resolution method
* output UTF-8 box drawing characters (with ASCII option)
* accept puzzle input with simplified synyax


bvdberg:
Nice indeed! The only issues is the 200Kb of output files. That makes the archive quite a lot bigger (I'm trying to keep it small and lean). Especially since it's only for the examples..

chqrlie:
Indeed, I did not do the math... I shall reduce this.

PR #551, opened 2026-07-25
Language: `import` accepts optional list of imported symbols

chqrlie:
* syntax is `import stdio { fopen, fclose, printf }`
* this helps readability and reduces symbol table lookups

### DIFF
--- a/analyser/module_analyser.c2
+++ b/analyser/module_analyser.c2
@@ -172,6 +172,9 @@ public fn void Analyser.checkModule(Analyser* ma, Module* mod) {
 
     if (ma.diags.hasErrors()) return;
 
+    mod.visitImports(Analyser.handleImportList, ma);
+    //if (ma.has_error) return;
+
     ma.collectTypeFunctions();
     if (ma.has_error) return;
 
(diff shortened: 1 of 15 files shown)


bvdberg:
I'm playing around with this. Some observations:
```
module foo;
public type Public struct { i32 x; }
type Private struct { i32 x; }
```

- import foo { Private }  will give an error: module 'foo' has no symbol 'Private' (it should say: symbol 'foo.Private' is not public
- Detection of ambiguous symbols still works with:
```
import foo { Public }
import bar local;

// Public a; // correct error
```

I'm not sure if I like the {} notation. I'm playing around with symbols:
```
        import stdio { printf }
        import stdio { * }      // old import local
        import stdio;
    OR
        import stdio [ printf, sprintf ];
        import stdio [ * ];
        import stdio;
    OR
        import stdio ( printf, sprintf );
        import stdio ( * );
        import stdio;
```

I like the [] and () also, except for the traling semi-colon....

For implementation, I expected that you used the InitListExpr to store the import-list. I think your version is a bit more compact in memory (as it is tail-allocated). We will need to update the refs plugin for this, so can a ref for 'printf' is generated.

bvdberg:
Wow this is getting fancy... The u32 ImportDecl.symlist_count causes the ImportDecl to grow by 8 bytes. In Decl there are still 17 bits in the flags left. That would easily fit any list a user could add. Then the size would remain 40 bytes :)

chqrlie:
> Wow this is getting fancy... The u32 ImportDecl.symlist_count causes the ImportDecl to grow by 8 bytes. In Decl there are still 17 bits in the flags left. That would easily fit any list a user could add. Then the size would remain 40 bytes :)

Good point, actually only 16 bits left in the `ImportDeclBits`, a nicely aligned 16-bit word.... Done!

bvdberg:
I found one more case we might need to think about:
```
import bar { Public as Public2 }
```
Where Public2 would also be a public struct of bar. So this hides bar.Public2. We might just allow this, or give a warning/error..

PR #553, opened 2026-07-28
Tester: substantial rewrite with simpler parser

chqrlie:
* simplified parser
* factorize keyword/tag parsers
* factorize Issues handlers
* improve filename output with full temporary file path in errors


bvdberg:
merged

PR #557, opened 2026-08-04
Github CI: split header-dump as a separate action

chqrlie:


bvdberg:
Is it correct that I only see stuff removed? Did you forget to add the new file? (if that was created)

chqrlie:
> Is it correct that I only see stuff removed? Did you forget to add the new file? (if that was created)

The file **[.github/workflows/header-dump.yml‎](https://github.com/c2lang/c2compiler/pull/557/changes#diff-e0ff02a47d5deb5c6db2b623238f832b8f65c8074ca2280f833f4387a3ad3068)** was added.  You might not see it because it is in a hidden directory :)

